PR #1, opened 2017-09-29
Create empty new ez_rubocop gem

tjwp:
# What did we change?

There are two commits here. The first was to create a new gem using:

`bundle gem ez_rubocop --no-coc --test=rspec`

The second was to customize the generated code. Some of this, e.g. the versioning approach, is based on similar work that I did previously: https://github.com/salsify/salsify_rubocop

This gem is public because eventually we'll want to use it to share rubocop configuration across both public and private gems.

The rubocop configuration in this gem will eventually be replaced by a shared configuration provided by this gem.

# Why are we doing this?

- To provide a place to Mark to distribute a custom cop that he is planning to develop.

- To start a conversation about how we build gems, as I expect we'll soon be creating a lot more.


lib/ez_rubocop/version.rb
```diff
@@ -0,0 +1,3 @@
+module EzRubocop
+  VERSION = "0.49.0".freeze
```
mjfaga: Do we want to use the most recent version here?

tjwp: I think we should start with matching what ez-rails is using, so that we can drop this into ez-rails once it has code to share. 

I'd rather burn down some of the existing offenses in ez-rails before introducing another rubocop update there.

AlexMorreale:
@wpeterson Tim and I had the same discussion offline

tjwp:
In our offline discussion, Alex suggested changing the prefix to `ezcater`. How do folks feel about standardizing on that? I think it makes sense to have a prefix that is unique to us.

I would argue that we stick with an underscore instead of a hyphen in this case, due to the convention of hyphens separating namespaces in the ruby gem world:
- `ezcater_rubocop` -> `EzcaterRubocop`
- `ezcater-rubocop` -> `Ezcater::Rubocop`

wpeterson:
I like `ezcater` as a prefix 👍 

I like gem names with `underscores` as a delimeter instead of dashes, since the gem name, repo name, and require name all align.  I feel strongly we should do this as you proposed.  [Reference: naming gems](http://guides.rubygems.org/name-your-gem/)

I think Gems should probably own a top-level Ruby namespace particular to that gem, preferring `EzcaterRubocop` to `Ezcater::Rubocop` which shares an `Ezcater` namespace, but I don't feel very strongly about this topic.

PR #2, opened 2017-10-03
Add FeatureFlag spec cop

mjfaga:
1. Add dependencies for writing specs for cops
2. Add support for pry gem for debugging
3. Add RspecRequireFeatureFlagMock Cop w/specs

### DIFF
--- a/ezcater_rubocop.gemspec
+++ b/ezcater_rubocop.gemspec
@@ -34,6 +34,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "bundler", "~> 1.15"
   spec.add_development_dependency "rake", "~> 10.0"
   spec.add_development_dependency "rspec", "~> 3.0"
+  spec.add_development_dependency "pry-byebug"
 
   spec.add_runtime_dependency "rubocop", "~> 0.49.1"
   spec.add_runtime_dependency "rubocop-rspec", "~> 1.15.1"
--- a/lib/rubocop/cop/ezcater/rspec_require_feature_flag_mock.rb
+++ b/lib/rubocop/cop/ezcater/rspec_require_feature_flag_mock.rb
@@ -1,0 +1,52 @@
+module RuboCop
+  module Cop
+    module Ezcater
+      # Enforce usage of our `mock_feature_flag` helpers over explicit usage
+      # of allow(FeatureFlag) mocking.
+      #
+      # @example
+      #
+      #   # good
+      #   mock_feature_flag("MyFeatureFlag", true)
+      #   mock_feature_flag("MyFeatureFlag", { user: current_user }, true)
+      #
+      #   # bad
+      #   allow(FeatureFlag).to receive(:is_active?).and_return(true)
+      #   allow(FeatureFlag).to receive(:is_active?).with("MyFeatureFlag").and_return(true)
+      #   allow(FeatureFlag).to receive(:is_active?).with("MyFeatureFlag", user: current_user).and_return(true)
+      class RspecRequireFeatureFlagMock < Cop
+        MSG = "Use the `mock_feature_flag` helper instead of mocking `allow(FeatureFlag)`".freeze
+
+        def_node_matcher :feature_flag_const?, <<~PATTERN
+          (const _ :FeatureFlag)
+        PATTERN
+
+        def on_const(node)
+          return unless feature_flag_const?(node)
+
+          # Walk to send node where method = :allow
+          match_node = node
+          match_node = match_node.parent while not_allow_send_node?(match_node.parent)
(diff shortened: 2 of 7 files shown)


ezcater_rubocop.gemspec
```diff
@@ -34,6 +34,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "bundler", "~> 1.15"
   spec.add_development_dependency "rake", "~> 10.0"
   spec.add_development_dependency "rspec", "~> 3.0"
+  spec.add_development_dependency "pry-byebug"
```
zoso10: doesn't bundler complain when a version isn't specified? not a huge deal since it's pry and dev dependency

lib/rubocop/cop/ezcater/rspec_require_feature_flag_mock.rb
```diff
@@ -0,0 +1,53 @@
...
+      #   mock_feature_flag("MyFeatureFlag", true)
+      #   mock_feature_flag("MyFeatureFlag", { user: current_user }, true)
+      #
+      #   # bad
+      #   allow(FeatureFlag).to receive(:is_active?).and_return(true)
+      #   allow(FeatureFlag).to receive(:is_active?).with("MyFeatureFlag").and_return(true)
+      #   allow(FeatureFlag).to receive(:is_active?).with("MyFeatureFlag", user: current_user).and_return(true)
+      class RspecRequireFeatureFlagMock < Cop
+        MSG = "Please use the `mock_feature_flag` helper over " \
+          "`allow(FeatureFlag)` directly.".freeze
```
tjwp: Re-word suggestion "Use the `mock_feature_flag` helper instead of mocking `FeatureFlag.is_active?`"?

lib/rubocop/cop/ezcater/rspec_require_feature_flag_mock.rb
```diff
@@ -0,0 +1,53 @@
...
+
+        def_node_matcher :mock_feature_flag_candidate?, <<~PATTERN
```
tjwp: Perhaps rename for exactly the node matcher is: `feature_flag_const?`?

spec/rubocop/cop/ezcater/rspec_require_feature_flag_mock_spec.rb
```diff
@@ -0,0 +1,35 @@
+require "spec_helper"
+
+RSpec.describe RuboCop::Cop::Ezcater::RspecRequireFeatureFlagMock, :config do
```
tjwp: Aside: how do you feel about requiring the `RSpec.` here?

zoso10: for some reason i thought rspec recommended one over the other? cant remember tho

mjfaga: I like this - we do this in sem-tools as well. You also made it part of the configuration in the gem, so I assumed you desired this as well. If not, we would need to remove `config.disable_monkey_patching!` from `spec_helper.rb`

ezcater_rubocop.gemspec
mjfaga: Nope... it just always pulls the latest when you update:
image

spec/rubocop/cop/ezcater/rspec_require_feature_flag_mock_spec.rb
tjwp: When you run `rspec install` now it includes the `disable_monkey_patching!` so I left it as is. 

Since it seems to be their recommendation, and it's a pretty minimal change, just being on the `describe`, I'm good with it. 👍

ezcater_rubocop.gemspec
tjwp: To add to that, for a gem the `Gemfile.lock` is not committed. It will be generated and used locally, but developers working on the gem may use different versions unless you constrain them here.

mjfaga: I'm ok with that for `pry`.... if you guys disagree, I'm happy to lock it to a version, but until there are issues with `pry`, I don't see any reason to force people onto a particular version. Let me know thoughts.

tjwp: I'm fine with no version restriction on pry.

zoso10: same

mjfaga:
[Prevent usage of `allow(FeatureFlag)` and force people to use mock helpers](https://trello.com/c/Vc247Yws/314-prevent-usage-of-allowfeatureflag-and-force-people-to-use-mock-helpers)

PR #3, opened 2017-10-03
Cop transfers

tjwp:
This change copies the `StyleDig` and `RspecDotNotSelfDot` cops from salsify_rubocop.

The code is copied in a separate commit, rubocop offenses based on this repo's config are corrected in a separate commit, and any other edits required to get the code working in this repo are a further commit.

There is a another commit to add default configuration, REAME, changelog, etc.

And a final round of fixing rubocop offenses.

I've manually tested this code with ez-rails and it seems to be working as expected. I'd like to release a 0.49.0.rc0 release candidate once this PR is merged, and then test that with ez-rails in CI. After that I'll make an official 0.49.0 release.

This project should be setup in Circle. My plan is to handle that in a separate PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,3 @@
 
 # rspec failure tracking
 .rspec_status
-
-# RubyMine files
-/.idea/*
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,6 @@
 
 # rspec failure tracking
 .rspec_status
+
+# RubyMine files
+/.idea/*
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # ezcater_rubocop
 
 # v0.49.0 (unreleased)
-- Initial release
-- [RequireFeatureFlagMock](https://github.com/ezcater/ezcater_rubocop/blob/9adbce126ba971b48a9944ef97d2ed6330342632/lib/rubocop/cop/ezcater/rspec_require_feature_flag_mock.rb) Cop
-- [RspecDotNotSelfDot](https://github.com/ezcater/ezcater_rubocop/blob/master/lib/rubocop/cop/ezcater/rspec_dot_not_self_dot.rb) Cop
-- [StyleDig](https://github.com/ezcater/ezcater_rubocop/blob/master/lib/rubocop/cop/ezcater/style_dig.rb) Cop
+- Initial release.
+- Add `Ezcater/RequireFeatureFlagMock` cop.
+- Add `Ezcater/RspecDotNotSelfDot` cop.
+- Add `Ezcater/StyleDig` cop.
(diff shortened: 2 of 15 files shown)


CHANGELOG.md
```diff
@@ -3,3 +3,5 @@
 # v0.49.0 (unreleased)
 - Initial release
 - [RequireFeatureFlagMock](https://github.com/ezcater/ezcater_rubocop/blob/9adbce126ba971b48a9944ef97d2ed6330342632/lib/rubocop/cop/ezcater/rspec_require_feature_flag_mock.rb) Cop
+- [RspecDotNotSelfDot](https://github.com/ezcater/ezcater_rubocop/blob/master/lib/rubocop/cop/ezcater/rspec_dot_not_self_dot.rb) Cop
```
mjfaga: I think we should like the the commit hear, not master, since master will change over time... thoughts?

CHANGELOG.md
```diff
@@ -3,3 +3,5 @@
...
+- [StyleDig](https://github.com/ezcater/ezcater_rubocop/blob/master/lib/rubocop/cop/ezcater/style_dig.rb) Cop
```
mjfaga: same - link to commit?

README.md
```diff
@@ -47,7 +47,9 @@ the latest compatible version each time that the MAJOR.MINOR version of `rubocop
 is updated.
 
 ## Custom Cops
+1. [RspecDotNotSelfDot](https://github.com/ezcater/ezcater_rubocop/blob/master/lib/rubocop/cop/ezcater/rspec_dot_not_self_dot.rb) - Enforce ".<class method>" instead of "self.<class method>" for example group description.
```
mjfaga: These links, on the other hand, should link to master to ensure they always link to the current version.... 👍

CHANGELOG.md
tjwp: Yes, I thought about that when I was adding this. I wouldn't usually bother including links in the CHANGELOG. Also, I was going to squash these commits before merging, so I don't actually know the commit in advance. Unless we want to go back and fill out the CHANGELOG.

What do you think about just removing the links here?

mjfaga: We can remove the links... perhaps I was getting ahead of myself when i did that....

.gitignore
```diff
@@ -10,3 +10,6 @@
 
 # rspec failure tracking
 .rspec_status
+
+# RubyMine files
+/.idea/*
```
AlexMorreale: shouldn't people just be putting this in their global .gitignore

tjwp: I don't feel too strongly on this either way. We have it in ez-rails (not the strongest argument, I know) and it seems like a good insurance policy against people accidentally committing these files if they don't have it in a global .gitignore.

tjwp: Removed this in the interest of getting these changes completed.

PR #5, opened 2017-10-04
Add CircleCI configuration

mjfaga:
1. Add base Circle CI configuration
2. Configure Rubocop to run 7 fail fast in circle
3. Resolve existing rubocop issues

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,6 +26,11 @@ jobs:
           paths:
             - vendor/bundle
 
+      # Run Rubocop
+      - run:
+          name: RuboCop
+          command: bundle exec rubocop
+
       # Run rspec in parallel
       - type: shell
         command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,6 +35,7 @@ jobs:
       - type: shell
         command: |
           bundle exec rspec --profile 10 \
+                            --format RspecJunitFormatter \
                             --out test_results/rspec.xml \
                             --format progress
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,40 @@
+version: 2
+jobs:
+  build:
+    parallelism: 1
+    working_directory: ~/ezcater
+    docker:
+      - image: circleci/ruby:2.3.3
+    steps:
+      - checkout
+
+      # Restore bundle cache
+      - restore_cache:
+          keys:
+            - ruby-cache-{{ .Branch }}-{{ checksum "Gemfile" }}-{{ checksum "ezcater_rubocop.gemspec" }}
+            - ruby-cache-{{ .Branch }}-
+            - ruby-cache-
+
+      # Bundle install dependencies
+      - run: gem install bundler --no-document
+      - run: bundle check --path=vendor/bundle || bundle install --path=vendor/bundle --jobs=4 --retry=3
+      - run: bundle clean --force
+
+      # Store bundle cache
+      - save_cache:
+          key: ruby-cache-{{ .Branch }}-{{ checksum "Gemfile" }}-{{ checksum "ezcater_rubocop.gemspec" }}
+          paths:
+            - vendor/bundle
+
+      # Run rspec in parallel
+      - type: shell
+        command: |
+          bundle exec rspec --profile 10 \
+                            --out test_results/rspec.xml \
+                            --format progress
+
+      # Save test results for timing analysis
+      - store_test_results:
+          path: test_results
+      - store_test_results:
+          path: tmp/capybara
--- a/ezcater_rubocop.gemspec
+++ b/ezcater_rubocop.gemspec
@@ -34,6 +34,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "bundler", "~> 1.15"
   spec.add_development_dependency "rake", "~> 10.0"
   spec.add_development_dependency "rspec", "~> 3.0"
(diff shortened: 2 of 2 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,45 @@
+version: 2
+jobs:
+  build:
+    parallelism: 1
+    working_directory: ~/ezcater
+    docker:
+      - image: circleci/ruby:2.4
```
tjwp: Can we start with 2.3.3 since that is what we are targeting with ez-rails right now? That may quickly switch to 2.4.1 once we put ez-rails support on a 0.49-release branch.

.circleci/config.yml
```diff
@@ -0,0 +1,45 @@
...
+    steps:
+      - checkout
+
+      # Restore bundle cache
+      - restore_cache:
+          keys:
+            - ruby-cache-{{ .Branch }}-{{ checksum "Gemfile" }}
```
tjwp: For a gem you also want to include `checksum X.gemspec` here. Probably not necessary to add additional fallbacks based on that.

.circleci/config.yml
```diff
@@ -0,0 +1,45 @@
...
+            - ruby-cache-{{ .Branch }}-
+            - ruby-cache-
+
+      # Bundle install dependencies
+      - run: bundle check --path=vendor/bundle || bundle install --path=vendor/bundle --jobs=4 --retry=3
+      - run: bundle clean --force
+
+      # Store bundle cache
+      - save_cache:
+          key: ruby-cache-{{ .Branch }}-{{ checksum "Gemfile" }}
```
tjwp: Same thing here about including the gemspec.

.circleci/config.yml
```diff
@@ -0,0 +1,45 @@
...
+          paths:
+            - vendor/bundle
+
+      # Run Rubocop
+      - run:
+          name: RuboCop
+          command: ruby bin/circle_rubocop.rb
```
tjwp: I think this is probably overkill and we should just always run `bundle exec rubocop` for the gem. 

Hopefully none of our gems get large enough to require this!

.circleci/config.yml
```diff
@@ -0,0 +1,45 @@
...
+
+      # Run rspec in parallel
+      - type: shell
+        command: |
+          bundle exec rspec --profile 10 \
+                            --out test_results/rspec.xml \
+                            --format progress \
+                            $(circleci tests glob "spec/**/*_spec.rb" | circleci tests split --split-by=timings)
```
tjwp: Similarly here, we probably don't need to worry about splitting, and probably will never reach the point of needing to parallelize these specs.

bin/circle_rubocop.rb
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env ruby
```
tjwp: Based on comments above I think this script can be removed. Also it might have other rubocop offenses with ruby 2.4.

.circleci/config.yml
mjfaga: yeah - we target different ruby versions in different apps. SemTools is on 2.4, so that is my bad. Will update to ezrails version for now....

mjfaga: Or do we think it's worth running this stuff on multiple ruby instances for the versions we currently target?

.circleci/config.yml
mjfaga: Why rely on us doing that locally when we can guarantee on the server? Seems like a "best practice" we should be encouraging to me. Thoughts?

.circleci/config.yml
mjfaga: I'm just trying to be consistent with the fallbacks we use in other configurations... didn't see a reason to stray from the template we already kind of have in place. That said, will replace the Gemfile check with the gemspec check.

bin/circle_rubocop.rb
mjfaga: See my comments above... left it for now.

.circleci/config.yml
tjwp: Does circle provide a simple way to do that? Since this is public, I'd ultimately rather get this switched to Travis where testing against multiple versions is easily supported. In fact one easy way to do it here might be to run [wwtd](https://github.com/grosser/wwtd)

.circleci/config.yml
tjwp: I'm not saying to rely on doing that locally. I'm just saying run `bundle exec rubocop` here instead of the `cirle_rubocop.rb` script.

.circleci/config.yml
```diff
@@ -0,0 +1,44 @@
...
+    working_directory: ~/ezcater
+    docker:
+      - image: circleci/ruby:2.3.3
+    steps:
+      - checkout
+
+      # Restore bundle cache
+      - restore_cache:
+          keys:
+            - ruby-cache-{{ .Branch }}-{{ checksum "ezcater_rubocop.gemspec" }}
```
tjwp: In general I think you want both the Gemfile and the gemspec here. We can leave off the Gemfile for now, but there are cases, for example if you need to point to github for a particularly dependency, where you need to do that in the Gemfile. Or use another tool like [appraisal](https://github.com/thoughtbot/appraisal)

mjfaga: ah - ok - i will add back in for completeness. Sorry about that!

.circleci/config.yml
mjfaga: 👍

PR #6, opened 2017-10-05
Fix constant targeting for Browser mock cop

mjfaga:
Fixing the Browser cop to ignore compound constants that happen to end in `Browser`/`EzBrowser`

### DIFF
--- a/lib/rubocop/cop/ezcater/rspec_require_browser_mock.rb
+++ b/lib/rubocop/cop/ezcater/rspec_require_browser_mock.rb
@@ -25,7 +25,7 @@ module RuboCop
         PATTERN
 
         def on_const(node)
-          return unless browser_const?(node)
+          return unless browser_const?(node) && node.descendants.empty?
 
           # Walk to send node where method = :allow
           match_node = node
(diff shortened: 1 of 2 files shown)


lib/rubocop/cop/ezcater/rspec_require_browser_mock.rb
```diff
@@ -25,7 +25,7 @@ class RspecRequireBrowserMock < Cop
         PATTERN
 
         def on_const(node)
-          return unless browser_const?(node)
+          return unless browser_const?(node) && node.descendants.empty?
```
tjwp: Just a question for my own understanding: for the const node, the descendants are the module consts that contain it? The node pattern relationships are always backwards from the the way I'd normally think about them!

mjfaga: Yeah, it seems backwards. In the cases this stuff was failing, the node that was being tagged was `s(:const,s(:const, nil, :SomeOtherClass), :Browser)`. In this case, `s(:const, nil, :SomeOtherClass)` is a descendant (or receiver) of `Browser`.

PR #8, opened 2017-10-08
Fix StyleDig to allow nested access in assignments

tjwp:
This PR is against the `v0.49-release`.

Previously this cop only allowed for `||=`, probably because that is what I encountered in the code base where this was used originally.

There is a separate PR for the master/v0.50 version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # ezcater_rubocop
 
+# v0.49.2
+- Do no apply `Ezcater/StyleDig` to assignments to with nested access.
+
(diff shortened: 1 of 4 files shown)


mjfaga:
Is there a reason we are opening separate branches/PRs vs. merging down once, cherry picking commits once the .50 version is approvied?

tjwp:
No real reason. I originally made the change for v0.50, but had just been dealing with the incompatibilities between versions, so I wanted to go ahead and create the v0.49 version.

In the future, I expect we'll just start with the v0.50 version and backport only if necessary.

PR #10, opened 2017-10-16
Add shared configuration

tjwp:
This adds some pretty minimal rubocop configuration that can be shared via this gem.

My thinking was to start small and we'll evolve this configuration as we hit things that we dislike, or we re-evaluate cops as we work to cleanup ez-rails.

What went into this initial configuration?

- Copied what we already had in this gem and configuration from ezcater-delivery.
- Disabled the complexity gems as I've never seen them add much value -- but happy to discuss keeping them.
- Disabled the `Style/SingleLineBlockParams` which was also disabled in sem-tools.

There is also some additional cleanup here as rubocop-rspec was not actually being applied previously.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ end
 Or to your gem's gemspec file:
 
 ```ruby
-spec.add_development_dependency "salsify_rubocop"
+spec.add_development_dependency "ezcater_rubocop"
 ```
 
 And then execute:
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # ezcater_rubocop
 
-ezCater custom cops, and eventually shared rubocop configuration.
+ezCater custom cops and shared RuboCop configuration.
 
 [RuboCop](https://github.com/bbatsov/rubocop) is a static code analyzer that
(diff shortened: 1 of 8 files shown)


spec/rubocop/cop/ezcater/rspec_require_browser_mock_spec.rb
```diff
@@ -46,7 +46,7 @@
     expect(cop.offenses).to be_empty
   end
 
-  it "accepts usage of the mock_ezcater_app helper with options" do
+  it "accepts usage of the mock_chrome_browser helper with options" do
```
mjfaga: Whoops! Thanks!

README.md
```diff
@@ -15,6 +15,12 @@ group :development do
 end
 ```
 
+Or to your gem's gemspec file:
+
+```ruby
+spec.add_development_dependency "salsify_rubocop"
```
mjfaga: did you mean ezcater_rubocop?

tjwp: Doh! Of course I did.

PR #11, opened 2017-10-17
Do not apply dig to ranges

tjwp:
Hit yet another edge case with this custom cop.

This change is against the `v0.49-release` branch.


tjwp:
Yeah, I'm about one edge case away from just abandoning this cop. I like the use of `dig` but may just be too hard to accurately detect.

PR #12, opened 2017-10-18
Add Ezcater/PrivateAttr cop

tjwp:
This change adds a new custom cop, `Ezcater/PrivateAttr`, that recommends the use of the methods provided by the [private_attr](https://github.com/jswanner/private_attr) gem. 

The implementation was based on the existing official `Lint/InaffectiveAccessModifier` cop: https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/cop/lint/ineffective_access_modifier.rb

The specs for the new cop use the newer `expect_offense`/`expect_no_offenses` syntax that the main repo is moving to.

This is something that we are trying out in newer applications, and we want something to remind people to use these methods from the gem.

Included with this change is a configuration update to raise the maximum allowed RSpec example size to 25, matching our method length limit.

### DIFF
--- a/lib/rubocop/cop/ezcater/private_attr.rb
+++ b/lib/rubocop/cop/ezcater/private_attr.rb
@@ -1,0 +1,104 @@
+module RuboCop
+  module Cop
+    module Ezcater
+      # This cop checks for the use of `attr_accessor`, `attr_reader`, `attr_writer`
+      # and `alias_method` after the access modifiers `private` and `protected`
+      # and requires the use of methods from the private_attr gem instead.
+      #
+      # @example
+      #
+      #   # bad
+      #
+      #   class C
+      #     private
+      #
+      #     attr_accessor :foo
+      #     attr_reader :bar
+      #     attr_writer :baz
+      #     alias_method :foobar, foo
+      #   end
+      #
+      #
+      # @example
+      #
+      #   # good
+      #
+      #   class C
+      #     private_attr_accessor :foo
+      #     private_attr_reader :bar
+      #     private_attr_writer :baz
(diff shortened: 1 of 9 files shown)


lib/rubocop/cop/ezcater/private_attr.rb
```diff
@@ -0,0 +1,104 @@
...
+      #
+      #   class C
+      #     private_attr_accessor :foo
+      #     private_attr_reader :bar
+      #     private_attr_writer :baz
+      #     private_alias_method :foobar, :foo
+      #
+      #     private
+      #
+      #     def shy
```
BrentWheeldon: Hah this is a fantastic example!

lib/rubocop/cop/ezcater/private_attr.rb
```diff
@@ -0,0 +1,104 @@
...
+
+          elsif node.kwbegin_type?
+            check_scope(node, current_visibility)
+          end
+
+          current_visibility
+        end
+
+        def add_access_affected_call(node, current_visibility)
+          @access_affected_calls[node.method_name] = [node, current_visibility]
```
BrentWheeldon: Rather than populating this ivar could this method just call `add_offense` directly? It might simplify things?

tjwp: Sure, my initial plan was just to lean on the wisdom of whoever implemented the cop that I cribbed from. But it's not clear that level of scoping is needed here. And the original author didn't write tests of the original cop that demonstrate why it is required. So I'm happy to simplify this, and see if any issues are uncovered in actual use.

PR #13, opened 2017-10-18
Add circle_rubocop.rb script

tjwp:
This change adds the `circle_rubocop.rb` script to the gem so that it can be shared between applications.

I already cut a pre-release with the version number included here to test this with `ezcater-identity`: https://circleci.com/gh/ezcater/ezcater-identity/30

There are a couple of changes in flight for this gem, so they will likely be combined, and conflicts resolved prior to releasing a new official version.

### DIFF
--- a/ezcater_rubocop.gemspec
+++ b/ezcater_rubocop.gemspec
@@ -22,9 +22,20 @@ Gem::Specification.new do |spec|
       "public gem pushes."
   end
 
+  excluded_files = %w[.circleci/config.yml
+                      .gitignore
+                      .rspec
+                      .rubocop.yml
+                      .ruby-gemset
+                      .ruby-version
+                      .travis.yml
+                      bin/console
+                      bin/setup
+                      Rakefile]
+
   spec.files = `git ls-files -z`.split("\x0").reject do |f|
     f.match(%r{^(test|spec|features)/})
-  end
+  end - excluded_files
   spec.bindir = "bin"
   spec.executables << "circle_rubocop.rb"
   spec.require_paths = ["lib"]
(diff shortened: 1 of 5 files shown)


ezcater_rubocop.gemspec
```diff
@@ -25,8 +25,8 @@ Gem::Specification.new do |spec|
   spec.files = `git ls-files -z`.split("\x0").reject do |f|
     f.match(%r{^(test|spec|features)/})
   end
-  spec.bindir        = "exe"
-  spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
+  spec.bindir = "bin"
```
zoso10: aversions to keeping project-specific binstubs and gem inclusion-specific binstubs separate?

zoso10: oh wait is this so that we can also use the `circle_rubocop.rb` in this project?

tjwp: I changed the `bindir` because it seemed to match the most recent recommendations for structuring gems. I was not actually planning to run this script for this gem's circle build.

I'd be okay with separating the contents of the current "bin" directory, but I'm not sure it is necessary.

The `setup` and `console` files do get packaged with the gem currently, but only the `circle_rubocop.rb` script is executable, and therefore easily executed from the installed gem.

Rather than splitting, the better approach may be to exclude the other "bin" files from being packaged?

zoso10: makes sense 👍 TIL that that's the direction the community is taking. i think this approach makes more sense anyway, at the end of the day they're all binstubs only expose the ones you want, rite?

tjwp:
Rebased and updated to exclude files (including `bin/setup` and `bin/console`) from the package.

PR #18, opened 2017-11-13
Bump version of rubocop gems.

BrentWheeldon:
I would like to bump the `rubocop` gem to get bbatsov/rubocop#4755 which prevents the `Rails/HasManyOrHasOneDependent` cop from failing if a `has_many` association has the `through` option specified.

I took the opportunity to bump `rubocop-rspec` just to keep up.

### DIFF
--- a/conf/rubocop.yml
+++ b/conf/rubocop.yml
@@ -31,6 +31,9 @@ Metrics/PerceivedComplexity:
 Rails:
   Enabled: false
 
+RSpec/ContextWording:
+  Enabled: false
+
 RSpec/ExampleLength:
   Max: 25
 
@@ -61,6 +64,10 @@ Style/PercentLiteralDelimiters:
 
 Style/SingleLineBlockParams:
   Enabled: false
+
+Style/StderrPuts:
+  Exclude:
+    - "bin/yarn"
 
 Style/StringLiterals:
   EnforcedStyle: double_quotes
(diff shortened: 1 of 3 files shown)


BrentWheeldon:
@tjwp Good call.

`ezcater-delivery` has 11 new offenses (2 from `Style/StderrPuts` (in a generated file - `bin/yarn`) and 9 from `RSpec/ContextWording`).

`ezcater-identity` has 6 new offences (the same 2 from `Style/StderrPuts` and 4 from `RSpec/ContextWording`).

I'm happy to take care of those when I bump the version of the this gem in those repos?

tjwp:
Thanks for checking.

I think we should disable `Style/StderrPuts` for `bin/yarn`. For the `RSpec/ContextWording` I saw that PR go through and thought it would be too restrictive. I think it limits you to context beginning with "with" or "when" by default? I think we should disable that cop entirely for now and style council can discuss, but my guess is that one is too restrictive/opinionated to enable.

BrentWheeldon:
Cool - should those updates go in `config/default.yml`?

tjwp:
Those should go into `conf/rubocop.yml` here.

BrentWheeldon:
Updated the defaults and confirmed that both `ezcater-delivery` and `ezcater-identity` are error-free.

PR #25, opened 2018-01-12
0.49.5: Gql error helpers Cop

mjfaga:
Enforces use of the helpers provided by `GQLErrors` instead of raising `GraphQL::ExecutionError` directly.

**Backwards compatibility release for ez-rails**

### DIFF
--- a/lib/rubocop/cop/ezcater/rspec_require_gql_error_helpers.rb
+++ b/lib/rubocop/cop/ezcater/rspec_require_gql_error_helpers.rb
@@ -1,0 +1,35 @@
+module RuboCop
+  module Cop
+    module Ezcater
+      # Enforce use of GQLErrors helpers instead of throwing
+      # GraphQL::ExecutionErrors directly
+      #
+      # @example
+      #
+      #   # good
+      #   GQLErrors.summary_error("An error occurred")
+      #   GQLErrors.request_error("You can't access this", 401)
+      #   GQLErrors.field_error("is invalid", :first_name, "First Name")
+      #   GQLErrors.field_errors_for(my_model, context)
+      #   GQLErrors.field_errors_for(my_model, context, summary_error: "An error occurred")
+      #   GQLErrors.field_errors_for(my_model, context, field_mapping: { first: :first_name })
+      #
+      #   # bad
+      #   GraphQL::ExecutionError.new("An error occurred")
+      #   GraphQL::ExecutionError.new("You can't access this", options: { status_code: 401 })
+      class RspecRequireGqlErrorHelpers < Cop
+        MSG = "Use the helpers provided by `GQLErrors` instead of raising `GraphQL::ExecutionError` directly.".freeze
+
+        def_node_matcher :graphql_const?, <<~PATTERN
+          (const (const _ :GraphQL) :ExecutionError)
+        PATTERN
+
+        def on_const(node)
+          return unless graphql_const?(node)
+
+          add_offense(node, :expression, MSG)
+        end
+      end
+    end
+  end
+end
--- a/spec/rubocop/cop/ezcater/rspec_require_gql_error_helpers_spec.rb
+++ b/spec/rubocop/cop/ezcater/rspec_require_gql_error_helpers_spec.rb
@@ -1,0 +1,35 @@
+require "spec_helper"
+
+RSpec.describe RuboCop::Cop::Ezcater::RspecRequireGqlErrorHelpers, :config do
+  subject(:cop) { described_class.new }
+
+  let(:error_message) { described_class::MSG }
+  let(:expected_source) { "GraphQL::ExecutionError" }
+
+  context "when attempting to directly use GraphQL::ExecutionError" do
+    it "registers an offense" do
+      source = "error.is_a? GraphQL::ExecutionError"
+      inspect_source(cop, source)
+      expect(cop.messages).to eq([error_message])
+      expect(cop.highlights).to eq([expected_source])
+    end
+
+    context "and instantiate a new instance" do
+      it "registers an offense" do
+        source = "GraphQL::ExecutionError.new(\"An error occurred\")"
+        inspect_source(cop, source)
+        expect(cop.messages).to eq([error_message])
+        expect(cop.highlights).to eq([expected_source])
+      end
+
+      context "and additional options are provided" do
+        it "registers an offense" do
+          source = "GraphQL::ExecutionError.new(\"An error occurred\", options: { status_code: 401 })"
+          inspect_source(cop, source)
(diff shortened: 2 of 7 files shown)


mjfaga:
Pulled this branch into ez-rails manually and tested that ez-rails is currently passing: https://circleci.com/gh/ezcater/ez-rails/58302

spec/rubocop/cop/ezcater/rspec_require_gql_error_helpers_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  it "accepts usage of the GQLErrors.field_error" do
+    inspect_source(cop, "GQLErrors.field_error(\"is invalid\", :first_name, \"First Name\")")
+    expect(cop.offenses).to be_empty
+  end
+
+  it "accepts usage of the GQLErrors.field_errors_for" do
+    inspect_source(cop, "GQLErrors.field_errors_for(my_model, context)")
+    expect(cop.offenses).to be_empty
+  end
```
tjwp: Since the cop is just concerned with `GraphQL::ExecutionError` do each of the `GQLErrors` examples add something?

lib/rubocop/cop/ezcater/rspec_require_gql_error_helpers.rb
```diff
@@ -0,0 +1,35 @@
...
+      #   GQLErrors.field_errors_for(my_model, context, field_mapping: { first: :first_name })
+      #
+      #   # bad
+      #   GraphQL::ExecutionError.new("An error occurred"
+      #   GraphQL::ExecutionError.new("You can't access this", options: {status_code: 401})
+      class RspecRequireGqlErrorHelpers < Cop
+        MSG = "Use the helpers provided by `GQLErrors` instead of raising `GraphQL::ExecutionError` directly.".freeze
+
+        def_node_matcher :graphql_const?, <<~PATTERN
+          (const (const _ :GraphQL) :ExecutionError)
```
tjwp: Should this check for calling `.new` on this class or should any reference to the class be forbidden?

For example, this cop would flag `rescue GraphQL::ExecutionError`.

lib/rubocop/cop/ezcater/rspec_require_gql_error_helpers.rb
```diff
@@ -0,0 +1,35 @@
...
+      #   # good
+      #   GQLErrors.summary_error("An error occurred")
+      #   GQLErrors.request_error("You can't access this", 401)
+      #   GQLErrors.field_error("is invalid", :first_name, "First Name")
+      #   GQLErrors.field_errors_for(my_model, context)
+      #   GQLErrors.field_errors_for(my_model, context, summary_error: "An error occurred")
+      #   GQLErrors.field_errors_for(my_model, context, field_mapping: { first: :first_name })
+      #
+      #   # bad
+      #   GraphQL::ExecutionError.new("An error occurred"
```
anishasmith: `)`?

spec/rubocop/cop/ezcater/rspec_require_gql_error_helpers_spec.rb
```diff
@@ -0,0 +1,37 @@
+require "spec_helper"
+
+RSpec.describe RuboCop::Cop::Ezcater::RspecRequireGqlErrorHelpers, :config do
+  subject(:cop) { described_class.new }
+
+  let(:error_message) { described_class::MSG }
+  let(:expected_source) { "GraphQL::ExecutionError" }
+
+  it "registers an offense when attempting to directly use GraphQL::ExecutionError with options" do
```
anishasmith: can we use context blocks here and below?

lib/rubocop/cop/ezcater/rspec_require_gql_error_helpers.rb
mjfaga: I think for now, I would like to start that way actually. I don't see a reason for us to be interacting with that class at all. I'd be curious on @christoomey's thoughts on this, but for now, I'd think I'd like to opt into leaving it. We can always loosen that restriction once we've come up with a legit use case for referencing the class. I can't think of any off the top of my head.

spec/rubocop/cop/ezcater/rspec_require_gql_error_helpers_spec.rb
mjfaga: Glad you commented on this. The short answer, I'm not sure it does. I debated whether or not I should add these, and landed on it being appropriate to validate that the "good" examples aren't impacted by this rule. That said, I can appreciate that what is being tested has nothing to do with the code that is written. I'll remove for now.

mjfaga:
@tjwp @anishasmith Updates pushed. Let me know if you have any other thoughts based on the updates + my comments!

lib/rubocop/cop/ezcater/rspec_require_gql_error_helpers.rb
tjwp: @mjfaga I'm fine with taking that approach for now. Will you be disabling this cop for the file where you define the `GQLErrors` helpers?

If we're taking this approach, could you add a test example where you're just referencing the class (without `.new`) to document that you're intentionally flagging all references to it not just initializing new instances?

mjfaga: Correct, this cop will be disabled in that 1 file.

spec/rubocop/cop/ezcater/rspec_require_gql_error_helpers_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  context "when attempting to directly use GraphQL::ExecutionError" do
+    it "registers an offense" do
+      source = "error.is_a? GraphQL::ExecutionError"
+      inspect_source(cop, source)
+      expect(cop.messages).to eq([error_message])
+      expect(cop.highlights).to eq([expected_source])
+    end
+
+    context "and instantiate a new instance" do
```
anishasmith: nitpick: "instantiating"

mjfaga:
[Add Rubocop rule to prevent use of `GraphQL:ExecutionError` in favor of helpers instead.](https://trello.com/c/6fyL4ZGq/271-add-rubocop-rule-to-prevent-use-of-graphqlexecutionerror-in-favor-of-helpers-instead)

mjfaga:
[Add Rubocop rule to prevent use of `GraphQL:ExecutionError` in favor of helpers instead.](https://trello.com/c/6fyL4ZGq/271-add-rubocop-rule-to-prevent-use-of-graphqlexecutionerror-in-favor-of-helpers-instead)

PR #28, opened 2018-01-16
Exclude config/environments/*.rb for Metrics/BlockLength

tjwp:
The blocks in these files for configuration tend to get lengthy, and we already have exclusions inside the files for several apps.

I was surprised that this is not already part of our shared configuration.


mjfaga:
I will add a decision denoting this change to our wiki.

PR #29, opened 2018-01-16
Add RailsConfiguration cop

tjwp:
Add a custom cop to ensure the use of `Rails.configuration` instead of `Rails.application.config`.


tjwp:
After this is reviewed I'll backport something that can be used for ez-rails.

mjfaga:
Hah - never mind - you answered the question above.

PR #32, opened 2018-02-06
Disable RSpec/LetSetup cop

zoso10:
we already decided this, but putting in PR in case i did anything fishy. once this is approved ill cut a release and update the style council decisions table in the wiki

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # ezcater_rubocop
 
+## v0.51.8 (unreleased)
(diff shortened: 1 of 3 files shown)


zoso10:
i've been doing releases in other gems as a separate commit straight to master with the only change being the removal of "(unreleased)" from the changelog. that way it leaves a clear mark in git history where a version was released without having to check tags

anishasmith:
@zoso10 I like that policy.

PR #33, opened 2018-02-11
Update to rubocop v0.52.1 and rubocop-rspec 1.22.2

tjwp:
This updates this gem to use the latest releases of the rubocop gems that it depends on.

The only offenses that I had to correct in the code for this gem were for `Style/FormatStringToken`.

I've excluded file in the `bin/` directory for the new cop `Style/MixinUsage`. I added this configuration for Rails apps because Rails-generated files in the `bin/` directory violate this cop, that flags `include` at the top-level.

These changes were tested out with:
- identity: https://github.com/ezcater/ezcater-identity/pull/112
- delivery: https://github.com/ezcater/ezcater-delivery/pull/218

I was tempted to require Ruby 2.4 with this release. I held off on that because the `tools` repo was recently updated to use this gem (at v0.51.8) but it is using 2.3.3.

I would be interested in eventually implementing the disabled-by-default `Layout/ClassStructure` but I've also held off on that for now as I expect it will require a separate conversation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # ezcater_rubocop
 
+## v0.52.0 (unreleased)
+- Update to rubocop v0.52.1 and rubocop-rspec v1.22.2.
+
 ## v0.51.8
 - Disable `RSpec/LetSetup` cop.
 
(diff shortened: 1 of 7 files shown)


mjfaga:
Also:
1. I like exploring `Layout/ClassStructure`, and think I'd be a fan of enabling that.
1. Regarding tools, I don't think there is anything in there that would be too concerning for jumping from 2.3.3 to 2.4, but am cool with making that a separate PR from the changes here. I think it's worth exploring this change next though.

PR #34, opened 2018-02-22
Allow staging as rails env

zoso10:
"staging" is not one of the default envs for this cop. so i added it
along with being explicit about the other 3. this is a bug fix so i only bumped the tiny/patch version number

(didnt tag tim so he can enjoy his vacation :smiley: )


zoso10:
TIL! didn't realize it was in the README :smiley:

PR #35, opened 2018-03-14
Disable the `Style/IfUnlessModifier` cop.

BrentWheeldon:
We felt that this rule wasn’t helping and wasn’t something we necessarily want to enforce.


anishasmith:
LGTM, missed by a second

PR #36, opened 2018-03-29
Exclude spec/integrations for RSpec/DescribeClass

tjwp:
## What did we change?

Excluded the `spec/integrations` for `RSpec/DescribeClass`.

## Why are we doing this?

Integration tests are another type of test where it does not make sense to require that the top-level describe is a class.

## How was it tested?
- [ ] Specs
- [ ] Locally


tjwp:
That's already there on line 43.

PR #39, opened 2018-04-04
Add RSpecMatchOrderedArrayNotEq cop

zoso10:


lib/rubocop/cop/ezcater/rspec_match_ordered_array_not_eq.rb
```diff
@@ -0,0 +1,44 @@
...
+      #   # good
+      #   expect(foo).to match_ordered_array([1, 2, 3])
+      #   expect(foo).to match_ordered_array [1, 2, 3]
+      #
+      #   # bad
+      #   expect(foo).to eq([1, 2, 3])
+      #   expect(foo).to eq [1, 2, 3]
+      class RspecMatchOrderedArrayNotEq < Cop
+        MATCH_ORDERED_ARRAY = "match_ordered_array".freeze
+        MSG = "Use the `match_ordered_array` matcher instead of `eq` when comparing collections".freeze
```
tjwp: Is it worth mentioning the `ezcater_matchers` gem when recommending this?

lib/rubocop/cop/ezcater/rspec_match_ordered_array_not_eq.rb
```diff
@@ -0,0 +1,44 @@
...
+      # @example
+      #
+      #   # good
+      #   expect(foo).to match_ordered_array([1, 2, 3])
+      #   expect(foo).to match_ordered_array [1, 2, 3]
+      #
+      #   # bad
+      #   expect(foo).to eq([1, 2, 3])
+      #   expect(foo).to eq [1, 2, 3]
+      class RspecMatchOrderedArrayNotEq < Cop
```
tjwp: Thoughts on calling this just `RspecMatchOrderedArray`?

lib/rubocop/cop/ezcater/rspec_match_ordered_array_not_eq.rb
anishasmith: ^^ I think we should. This could be a confusing google, otherwise.

spec/rubocop/cop/ezcater/rspec_match_ordered_array_not_eq_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  it "accepts usage of `eq` without an array" do
+    source = "expect(foo).to eq(1, 'foo', :baz, nil)"
+
+    inspect_source(source)
+
+    expect(cop.offenses).to be_empty
+  end
+
+  it "corrects when using `eq` matcher for arrays" do
```
anishasmith: 🎊

lib/rubocop/cop/ezcater/rspec_match_ordered_array_not_eq.rb
zoso10: totally!

lib/rubocop/cop/ezcater/rspec_match_ordered_array_not_eq.rb
zoso10: sure, i was following a convention we were already using with `Ezcater/RspecDotNotSelfDot`

zoso10: maybe "convention" isnt the proper word here, but I was following the naming of another that I saw.

tjwp: Yeah, that was a weird one where just calling it `RSpecDot` didn't make a lot of sense!

tjwp:
LGTM

PR #44, opened 2018-07-02
Add a cop to enforce response status assertion style

BrentWheeldon:
## What did we change?

- Add a cop to enforce response status assertion style
- Enable an existing cop to enforce response status assertion style

## Why are we doing this?

We decided that we'd like consistency here, and think the enforced style provides more clarity.

## How was it tested?
- [X] Specs
- [X] Locally


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # ezcater_rubocop
 
+## v0.57.1
+- Add `Ezcater/RspecRequireHttpStatusMatcher` cop.
+- Enable `RSpec/Rails/HttpStatus` cop and enforce symbols.
```
tjwp: Is `RSpec/Rails/HttpStatus` enabled here or `Rails/HttpStatus`?

lib/rubocop/cop/ezcater/rspec_require_http_status_matcher.rb
```diff
@@ -0,0 +1,35 @@
...
+      #
+      #   # good
+      #   expect(response).to be_created
+      #   expect(response).to be_bad_request
+      #
+      #   # bad
+      #   expect(response).to have_http_status 201
+      #   expect(response).to have_http_status 400
+      class RspecRequireHttpStatusMatcher < Cop
+        MSG = "Use the `have_http_status` matcher, like `expect(response).to have_http_status :bad_request`, "\
```
tjwp: Are the good and bad examples above reversed? This is recommended what is listed as "bad" above?

spec/rubocop/cop/ezcater/rspec_require_http_status_matcher_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+
+RSpec.describe RuboCop::Cop::Ezcater::RspecRequireHttpStatusMatcher do
+  subject(:cop) { described_class.new }
+
+  it "accepts usage of the HTTP status matcher" do
+    inspect_source("expect(response).to have_http_status :bad_request")
+    expect(cop.offenses).to be_empty
+  end
+
+  it "registeres an offence when attempting to assert directly on the status" do
```
tjwp: "registers"

spec/rubocop/cop/ezcater/rspec_require_http_status_matcher_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+    inspect_source("expect(response.code).to eq 400")
+    expect(cop.messages).to match_array(["Use the `have_http_status` matcher, like "\
+      "`expect(response).to have_http_status :bad_request`, rather than `expect(response.code).to eq 400`",])
+  end
+end
```
tjwp: Should there be a test case for `expect(response).to be_bad_request`?

BrentWheeldon: I wasn't planning on banning that style since it seemed non-trivial and, while not our preferred syntax, still provides more documentation value than asserting around magic integers.

BrentWheeldon: @bjackson2 and @mjfaga - do y'all think it's worth banning the `expect(response).to be_bad_request` style, too?

bjackson2: I'm OK with the `be_foo` style too, we use that style in a buncha places

mjfaga: I'm fine with leaving that in place. I think the goal/scope here is to avoid explicit status codes (`400`, `401`) in our specs and write more readable specs (`:bad_request`, `:unauthorized`).

PR #48, opened 2018-07-11
Configure `Rspec/MessageExpectation` with `allow`

mjfaga:
## What did we change?
Configure `Rspec/MessageExpectation` with the `allow` style.

## Why are we doing this?
Style council decision

## How was it tested?
- [ ] Specs
- [ ] Locally


mjfaga:
@brandonshar-ezcater correct. That is actually what it enforces.

PR #49, opened 2018-07-22
Update to rubocop v0.58.1

tjwp:
## What did we change?

Updated to the latest rubocop release.

Enabled instance variable naming cop now that configuration is available.

## Why are we doing this?

To keep up with rubocop releases.

This updated was tested against:
- ezcater_kafka
- ezcater-identity
- ezcater-sms-service
- ezcater-delivery

Only ezcater-delivery had a new offense, shown here: https://github.com/ezcater/ezcater-delivery/pull/621

## How was it tested?
- [x] Specs
- [ ] Locally

### DIFF
--- a/lib/ezcater_rubocop/version.rb
+++ b/lib/ezcater_rubocop/version.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module EzcaterRubocop
-  VERSION = "0.57.4"
+  VERSION = "0.58.0.rc0"
 end
--- a/lib/ezcater_rubocop/version.rb
+++ b/lib/ezcater_rubocop/version.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module EzcaterRubocop
-  VERSION = "0.58.0.rc0"
+  VERSION = "0.58.0.rc1"
 end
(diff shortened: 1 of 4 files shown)


tjwp:
My preference would be for this configuration, but ultimately this is a question for the style council to decide.

This is the first place that I've worked that has used the underscore convention for instance variables. My understanding is that there is semantic difference implied by the presence of the underscore. Without the underscore direct access to the instance variable is permissible, but with the underscore it should only be used via an accessor.

I can leave this PR open until the style council is able to discuss it.

BrentWheeldon:
Added to the [agenda](https://ezcater.atlassian.net/wiki/spaces/POL/pages/680657215/2018-07-23+Style+Council+Sync+notes) for today's sync.

tjwp:
Updated with required configuration.

lib/ezcater_rubocop/version.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module EzcaterRubocop
-  VERSION = "0.57.4"
+  VERSION = "0.58.0.rc1"
```
mjfaga: Do we still need the `rc1` here?

tjwp: I'll remove it before making the actual release. This was so that I could cut a new version with the latest configuration change and test it in other code bases.

PR #54, opened 2018-10-15
Turn off the RSpec/MessageExpectation cop

agrobbin:
The Style Council decided at the [10/15 meeting](https://ezcater.atlassian.net/wiki/spaces/POL/pages/757991514/2018-10-15+Style+Council+Sync+notes) to be more permissive of this syntax.

### DIFF
--- a/conf/rubocop.yml
+++ b/conf/rubocop.yml
@@ -62,10 +62,6 @@ RSpec/ExampleLength:
 RSpec/LetSetup:
   Enabled: false
 
-RSpec/MessageExpectation:
-  Enabled: true
-  EnforcedStyle: allow
-
 RSpec/MultipleExpectations:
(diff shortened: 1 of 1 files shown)


tjwp:
I'm a little concerned to see this configuration flapping. According to https://ezcater.atlassian.net/wiki/spaces/POL/pages/141131780/Style+Council+Decisions the current config was only introduced 7/2/2018 and code bases were updated to conform.

agrobbin:
Ohh that's interesting @tjwp, we didn't realize that when we were discussing this earlier today! That could definitely have influenced our decision. This might also point to an area of communication we need to improve, in terms of making sure future Style Council decisions take into account whether the item has already been adjudicated, along with how much past precedent factors into the discussion.

PR #56, opened 2018-11-05
Update `Metrics/BlockLength` to exclude `app/graphql/**/*.rb`

kevinmook:
## What did we change?
Update `Metrics/BlockLength` to exclude `app/graphql/**/*.rb`

## Why are we doing this?
The style council determined this rule isn't helpful for graphql types, given that graphql types are defined within a block. See https://ezcater.atlassian.net/wiki/spaces/POL/pages/786334408/2018-11-05+Style+Council+Sync+notes

## How was it tested?
- [x] Specs
- [x] Locally

### DIFF
--- a/conf/rubocop.yml
+++ b/conf/rubocop.yml
@@ -16,9 +16,6 @@ Metrics/BlockLength:
   Exclude:
   - "*.gemspec"
   - "spec/**/*.rb"
-  - "lib/tasks/**/*.rake"
-  - "config/environments/*.rb"
-  - "app/graphql/**/*.rb"
 
 Metrics/CyclomaticComplexity:
   Enabled: false
@@ -133,4 +130,3 @@ Style/TrailingCommaInArrayLiteral:
 
 Style/TrailingCommaInHashLiteral:
   EnforcedStyleForMultiline: consistent_comma
-
--- a/conf/rubocop.yml
+++ b/conf/rubocop.yml
@@ -18,6 +18,7 @@ Metrics/BlockLength:
   - "spec/**/*.rb"
   - "lib/tasks/**/*.rake"
   - "config/environments/*.rb"
+  - "app/graphql/**/*.rb"
 
 Metrics/CyclomaticComplexity:
   Enabled: false
--- a/conf/rubocop_rails.yml
+++ b/conf/rubocop_rails.yml
@@ -1,14 +1,13 @@
 inherit_from:
   - ../conf/rubocop.yml
 
-inherit_mode:
-  merge:
-    - Exclude
-
 AllCops:
   TargetRailsVersion: 5.1
 
 Metrics/BlockLength:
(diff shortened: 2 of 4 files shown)


kevinmook:
I figured this was a better place to update the rule, but let me know if this would be better to have live in ezRails given that there will be further discussions on this topic.

conf/rubocop.yml
```diff
@@ -18,6 +18,7 @@ Metrics/BlockLength:
   - "spec/**/*.rb"
   - "lib/tasks/**/*.rake"
   - "config/environments/*.rb"
+  - "app/graphql/**/*.rb"
```
BrentWheeldon: I wonder if we'd want to put this in the [rails specific conf](https://github.com/ezcater/ezcater_rubocop/blob/master/conf/rubocop_rails.yml)?

agrobbin: I think it's funny how this has become reverse-alphabetized!

lib/ezcater_rubocop/version.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module EzcaterRubocop
-  VERSION = "0.58.3"
+  VERSION = "0.58.4"
```
agrobbin: I lean away from including version bumps directly in PRs, instead preferring to commit the version bump change directly to the releasing branch after it has been merged. This allows for Git history to be consistent when 1 change is released on its own and when multiple changes are released together, as well as limiting the PR content to code that is purely related to the change.

conf/rubocop.yml
agrobbin: Unrelated to this change, but the `config/environments/*.rb` and `lib/tasks/**.*rake` exclusions should also probably be moved to the Rails-specific config file.

kevinmook:
@agrobbin @BrentWheeldon I updated this to move the BlockLength rules specific to rails into rubocop_rails.yml. I'm not sure if this would warrant a bigger version bump (I'm not sure how this is used), but let me know if there's anything else I need to do.

I did test this against ezRails master and the linter comes up all green

conf/rubocop_rails.yml
```diff
@@ -1,9 +1,19 @@
 inherit_from:
   - ../conf/rubocop.yml
 
+inherit_mode:
+  merge:
+    - Exclude
```
agrobbin: Is it possible to have this apply to only specific rules, so we don't suddenly change this across the board?

agrobbin: Awesome!!

lib/ezcater_rubocop/version.rb
tjwp: This flip side of this is that I like to see what kind of version bump the author intends to make, but hopefully that can be reflected in the CHANGELOG (as it is here).

PR #61, opened 2019-01-21
Release v0.61.1

wpeterson:
## What did we change?

* Update release notes in `CHANGELOG`
* Bump gem version to v. `0.61.1`
* Add Versioning section to `CHANGELOG` 

## Why are we doing this?

Release unpublished style rule changes.  Add versioning note to `CHANGELOG` to aid in versioning releases and reading the notes.

## How was it tested?
- [x] Specs
- [ ] Locally

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # ezcater_rubocop
 
+## v0.61.1
+- `Layout/IndentHash` enforces consistent style
+- `Layout/IndentArray` enforces consistent style
+
 ## v0.61.0
 - Update to `rubocop` v0.61.1.
 - Update to `rubocop-rspec` v1.30.1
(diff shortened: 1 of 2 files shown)


tjwp:
To date the minor version of this gem has only bumped with updates to the underlying rubocop library: https://github.com/ezcater/ezcater_rubocop#versioning

wpeterson:
@tjwp Thanks, sorry I missed this.  Would the appropriate version bump then be to `0.61.1` (ie increment patch version?)

tjwp:
Yes, `0.61.1` would be the next appropriate version.

FWIW, I'm not tied to keeping this versioning scheme. Its something that I used in the past and it worked well then, but it now seems to have tripped up multiple people here.

rubocop is working towards a 1.0 release and adopting semver after that. Perhaps at that point it makes sense to revisit the versioning for this gem.

PR #62, opened 2019-02-19
Removes EzCater/PrivateAttr custom cop

bjackson2:
## Why are we doing this?
The style council decided to remove this cop and go back to default Ruby syntax.


bjackson2:
Not in this codebase or in ez-rails... it's possible that other codebases may reference it directly. Normally I'd say it would be worthy of a minor version bump, but I know we're trying to stay in-line with Rubocop versions.

bjackson2:
Yah I was curious about this convention as well. We'll discuss next meeting

PR #64, opened 2019-02-27
Update language around versioning, bump v1.0.0

EliseFitz15:
## What did we change?
Bumping to v1.0.0 to start taking advantage of semantic versioning.

## Why are we doing this?
Easier to maintain our own appropriate versioning. Discussion in #62 for reference. 

## How was it tested?
- [ ] Specs
- [ ] Locally

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,15 @@
 
 # Versioning
 
-NB: This gem is versioned based on the `MAJOR`.`MINOR` version of rubocop. The first release of the
+ezCater is moving onto its own [Semantic Versioning](https://semver.org/) scheme starting with v1.0.0. All version bumps moving forward should increment using `MAJOR.MINOR.PATCH` based on changes.
+
+Prior to v1.0.0 this gem was versioned based on the `MAJOR`.`MINOR` version of RuboCop. The first release of the
 ezcater_rubocop gem was `v0.49.0`.
+
+## v1.0.0
+
+- Begin using Semantic Versioning
+- Delete `Ezcater/PrivateAttr`
 
 ## v0.61.1
 - `Layout/IndentHash` enforces consistent style
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 # Versioning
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,54 +2,74 @@
 
 # Versioning
 
-NB: This gem is versioned based on the `MAJOR`.`MINOR` version of rubocop. The first release of the
+ezCater is using [Semantic Versioning](https://semver.org/) moving forward with v1.0.0. All version bumps moving forward should increment using `MAJOR.MINOR.PATCH` based on changes.
```
bjackson2: 🎱 thoughts on saying something like 'moving onto its own semantic versioning scheme' or something similar? some wording that makes it more pronounced that 'we were doing rubocop's versions, now they are ours'

CHANGELOG.md
```diff
@@ -2,9 +2,16 @@
 
 # Versioning
 
-NB: This gem is versioned based on the `MAJOR`.`MINOR` version of rubocop. The first release of the
+ezCater is moving onto its own [Semantic Versioning](https://semver.org/) scheme starting with v1.0.0. All version bumps moving forward should increment using `MAJOR.MINOR.PATCH` based on changes.
```
wpeterson: nitpick: perhaps `ezcater_rubycop` or `This gem` vs. `ezCater` as the subject here?

README.md
```diff
@@ -70,17 +70,7 @@ running rubocop.
...
-
-The patch version for this gem does _not_ correspond to the patch version of
-`rubocop`. The patch version for this gem will change any time that one of its
-configurations is modified _or_ its dependency on `rubocop` is changed to require
-a different patch version.
-
-This gem also includes a dependency on `rubocop-rspec` that will be updated to
-the latest compatible version each time that the MAJOR.MINOR version of `rubocop`
-is updated.
+ezCater is using [Semantic Versioning](https://semver.org/). All version bumps should increment using `MAJOR.MINOR.PATCH` based on changes.
```
wpeterson: nitpick: perhaps `ezcater_rubycop` or `This gem` vs. `ezCater` as the subject here?

EliseFitz15: Yup good call! Will update in both.

EliseFitz15:
Good catch @wpeterson! Somehow missed checking in the version file 🙃 ! Let me know if this is good to go!

PR #65, opened 2019-03-06
Disable the Rails/HasAndBelongsToMany cop

agrobbin:
## What did we change?

Disable the `Rails/HasAndBelongsToMany` cop.

## Why are we doing this?

Rails' own [guides](https://guides.rubyonrails.org/association_basics.html#choosing-between-has-many-through-and-has-and-belongs-to-many) still recommend `has_and_belongs_to_many` in simple scenarios.

## How was it tested?
- [x] Specs
- [x] Locally


bjackson2:
We can discuss a new release at the meeting this coming week

wpeterson:
Posthumous LGTM

PR #70, opened 2019-03-13
Disable Style/DoubleNegation

wpeterson:
## What did we change?

* Disable `Style/DoubleNegation` cop
* Release `1.0.1`

## Why are we doing this?

Implements a decision we made in February I've now made real.

## How was it tested?
- [x] Specs
- [x] Locally


wpeterson:
```
[master ~/src/ezcater_rubocop]$> b rake release
ezcater_rubocop 1.0.1 built to pkg/ezcater_rubocop-1.0.1.gem.
Tagged v1.0.1.
Pushed git commits and tags.
Pushed ezcater_rubocop 1.0.1 to https://rubygems.org
```

PR #83, opened 2019-07-25
Add a cop to catch unwanted Rails.env use

aprescott:
## What did we change?

Added a cop to catch `Rails.env` references outside of `config/`.

## Why are we doing this?

Outside of `config/`, we want to avoid inspecting `Rails.env` to define
behavior. A couple of exceptions:

* Using `Rails.env`'s value for metadata-like purposes is permitted. For example, if some logging includes `Rails.env`, that's allowed.
* `db/` is ignored to allow dev-only migrations, which is a common existing pattern.

## How was it tested?
- [x] Specs
- [x] Locally

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -2,6 +2,14 @@ Ezcater/RailsConfiguration:
   Description: 'Enforce the use of `Rails.configuration` instead of `Rails.application.config`.'
   Enabled: true
 
+Ezcater/RailsEnv:
+  Description: 'Enforce the use of `Rails.configuration.x.<foo>` instead of `Rails.env` checks.'
+  Enabled: true
+  Exclude:
+    - "config/**/*"
+    - "spec/rails_helper.rb"
+    - "db/**/*"
+
 Ezcater/RspecDotNotSelfDot:
   Description: 'Enforce ".<class method>" instead of "self.<class method>" for example group description.'
   Enabled: true
--- a/lib/rubocop/cop/ezcater/rails_env.rb
+++ b/lib/rubocop/cop/ezcater/rails_env.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Ezcater
+      # Use the `Rails.configuration.x` namespace for configuration backed by
+      # environment variables, rather than inspecting `Rails.env` directly.
+      # Centralizing application configuration helps avoid scattering it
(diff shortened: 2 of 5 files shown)


lib/rubocop/cop/ezcater/rails_env.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Ezcater
+      class RailsEnv < Cop
```
aprescott: I was trying to avoid a very verbose name, but I'm open to better naming if this feels too generic.

tjwp: The name is fine with me.

config/default.yml
```diff
@@ -2,6 +2,14 @@ Ezcater/RailsConfiguration:
   Description: 'Enforce the use of `Rails.configuration` instead of `Rails.application.config`.'
   Enabled: true
 
+Ezcater/RailsEnv:
+  Description: 'Enforce the use of `Rails.configuration.x.<foo>` instead of `Rails.env` checks.'
+  Enabled: true
+  Exclude:
+    - "config/**/*"
+    - "spec/**/*"
```
agrobbin: Where do we anticipate this being used in `spec/**/*`? The only place I see it used that I think makes sense is in `spec/rails_helper.rb` (to avoid running tests in a production environment). Would it be possible to only exclude that file instead of all spec files?

lib/rubocop/cop/ezcater/rails_env.rb
agrobbin: Also fine with me!

config/default.yml
aprescott: I had this same question bouncing around my head for a while. `spec/rails_helper.rb` is indeed the only place it showed up under `spec/`, so I debated excluding just that file. I then started thinking about why people would inspect `Rails.env` in `spec/` but not `app/`, and could only come up with fairly unrealistic, contrived situations where we'd want to block the use, so I went with `spec/**/*`.

Thinking about it more though, in the absence of a sensible permitted use under `spec/` it's probably better to target the one file. It doesn't stop someone doing `allow(Rails.env).to receive(...)`, but the checks against `app/` would prevent those tests from applying anyway.

Will amend! 👍

aprescott:
Merging this but I'll hold off on a new release while I work on a similar `ENV` cop.

PR #84, opened 2019-07-29
Lint against ENV references outside of config/

aprescott:
## What did we change?

Similar to the `RailsEnv` cop (#83), but for `ENV` references.

## Why are we doing this?

To help us follow the decided env var best practices.

## How was it tested?
- [x] Specs
- [x] Locally

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -3,9 +3,18 @@ Ezcater/RailsConfiguration:
   Enabled: true
 
 Ezcater/RailsEnv:
-  Description: 'Enforce the use of `Rails.configuration.x.<foo>` instead of `Rails.env` checks.'
+  Description: 'Enforce the use of `Rails.configuration.x.<foo>` instead of checking `Rails.env`.'
   Enabled: true
   Exclude:
+    - "config/**/*"
+    - "spec/rails_helper.rb"
+    - "db/**/*"
+
+Ezcater/DirectEnvCheck:
+  Description: 'Enforce the use of `Rails.configuration.x.<foo>` instead of checking `ENV`.'
+  Enabled: true
+  Exclude:
+    - "bin/**/*"
     - "config/**/*"
     - "spec/rails_helper.rb"
     - "db/**/*"
--- a/config/default.yml
+++ b/config/default.yml
@@ -4,20 +4,11 @@ Ezcater/RailsConfiguration:
 
 Ezcater/RailsEnv:
   Description: 'Enforce the use of `Rails.configuration.x.<foo>` instead of checking `Rails.env`.'
-  Enabled: true
-  Exclude:
-    - "config/**/*"
-    - "spec/rails_helper.rb"
-    - "db/**/*"
+  Enabled: false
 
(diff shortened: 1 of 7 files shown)


lib/rubocop/cop/ezcater/direct_env_checks.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Ezcater
+      # Use the `Rails.configuration.x` namespace for configuration backed by
+      # environment variables, rather than inspecting `ENV` directly.
+      #
+      # Centralizing application configuration helps avoid scattering it
```
tjwp: 🎱 "Centralizing ... avoids scattering" reads like a bit of a tautology.

Maybe this line doesn't apply to this cop. Instead something like:

Restricting environment variables references to be within application configuration makes it more obvious which environment variables an applications relies upon.

config/default.yml
```diff
@@ -3,13 +3,22 @@ Ezcater/RailsConfiguration:
...
 Ezcater/RailsEnv:
-  Description: 'Enforce the use of `Rails.configuration.x.<foo>` instead of `Rails.env` checks.'
+  Description: 'Enforce the use of `Rails.configuration.x.<foo>` instead of checking `Rails.env`.'
   Enabled: true
   Exclude:
     - "config/**/*"
     - "spec/rails_helper.rb"
     - "db/**/*"
 
+Ezcater/DirectEnvChecks:
```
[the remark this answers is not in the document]
aprescott: Good catch, I'll singularize it. :namingishard:

lib/rubocop/cop/ezcater/direct_env_checks.rb
aprescott: Seems reasonable to me. 👍

aprescott:
> disabling by default, and then enabling in the Rails-specific `conf/rubocop_rails.yml`.

Went with this approach, in the absence of any clear alternative.

aprescott:
I think this is outside of the scope of this PR, but I could imagine us wanting to enforce some form of consistent `ENV` use in non-Rails apps, which might mean `Ezcater/DirectEnvCheck` applies more broadly, but we'd have to agree that non-Rails apps have, say, a common `config/` directory. Wider discussion, for sure.

I also debated calling this cop something like `Ezcater/RailsDirectEnvCheck`, but it starts to get confusing with `Ezcater/RailsEnv` and `Ezcater/RailsConfiguration` cops.

tjwp:
@aprescott My concern about enforcing this outside Rails was more for gems than non-Rails apps. We don't have a lot of non-Rails apps and I don't see that as a growth area for us, due to the lack of standardization.

tjwp:
@aprescott All devs should now have Github admin permissions on this repo. Let's see if that fixes your issue with CC.

aprescott:
@tjwp doesn't seem to have. I'm going to just ignore the issues for now and continue ahead.

PR #89, opened 2019-11-15
Add a cop to require snake_case fields in graphQL definitions

tvec:
## What did we change?
* Add a cop so we can follow our style guide seen here: https://git.io/JeofW.
  * the enforced style will be `:snake_case`
  * Largely inspired by the [method_name cop](https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/cop/naming/method_name.rb)

## Why are we doing this?
* Enforce Style GraphQL
* My real motivation is this: https://github.com/rmosolgo/graphql-ruby/blob/master/CHANGELOG.md#breaking-changes-6
  * When attempting to update the gem, many fields broke because the resolver method was not defined.
  * Because `snake_case` is idiomatic in ruby especially for methods, it made sense to enforce this to hopefully make it easier to port over and add methods

## NOTE
* A consequence of this PR will be to regenerate the todo file (at least in ezRails)
  * via `rubocop --auto-gen-config --exclude-limit 10000`
  * This allows us some time to go and clean up the field names because I didn't want to blindly snake case things and break schemas 😃
  

## PICS:
(taken on https://github.com/ezcater/ezcater_rubocop/pull/89/commits/7684ae69fe0de164cb3d882716ae0949b22f331c)
Screen Shot 2019-11-15 at 11 57 13 AM

## How was it tested?
- [x] Specs
- [x] Locally

```ruby
# manually test by updating gem source
gem "ezcater_rubocop", require: false, :path => '~/dev/ezcater/ezcater_rubocop/'
```

```bash
# in your favorite console
$ rubocop -d

# add the failed cases from above to specs
```

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -1,3 +1,12 @@
+Ezcater/GraphqlFieldsNaming:
+  EnforcedStyle: snake_case
+  Enabled: true
+  SupportedStyles:
+    - snake_case
+    - camelCase
+  Include:
+    - 'app/graphql/**/*.rb'
+
 Ezcater/RailsConfiguration:
   Description: 'Enforce the use of `Rails.configuration` instead of `Rails.application.config`.'
   Enabled: true
--- a/lib/rubocop/cop/ezcater/graphql_fields_naming.rb
+++ b/lib/rubocop/cop/ezcater/graphql_fields_naming.rb
@@ -1,0 +1,74 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Ezcater
+      # This cop makes sure that GraphQL fields & arguments match the supported style
+      # https://git.io/JeofW
+      #
+      # The cop also ignores when users provide a :camelize option, because
+      # the user is manually requesting to override the value
+      #
+      # @example
+      #   # bad
+      #   field :fooBar, ID, null: false
+      #   argument :barBaz, ID, required: true
+      #
+      #   # good
+      #   field :foo_bar, ID, null: true
+      #   field :foo_bar, ID, null: true do
+      #     argument :bar_baz, ID, required: true
+      #   end
+      #
+      #   field :fooBar, ID, null: true, camelize: true
+      #
+      class GraphqlFieldsNaming < Cop
+        include ConfigurableNaming
+
+        MSG = "Use %<style>s for GraphQL fields & arguments names. " \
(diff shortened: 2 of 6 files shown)


lib/rubocop/cop/ezcater/graphql_fields_naming.rb
```diff
@@ -0,0 +1,74 @@
...
+      class GraphqlFieldsNaming < Cop
+        include ConfigurableNaming
+
+        MSG = "Use %<style>s for GraphQL fields & arguments names. " \
+              "See https://git.io/JeofW for our guide. " \
+              "This can be overridden with camelize."
+        FIELD_ADDING_METHODS = %i(field argument).freeze
+        PROCESSABLE_TYPES = %i(sym string).freeze
+
+        def on_send(node)
```
tvec: The method definition for fields can be seen here: https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/member/has_fields.rb#L9-L16

tvec:
Sadly, it does not 😢 even though I wish I could.

The reason why I don't think I can add that option is because it can technically break code:

```ruby
# BEFORE
field :coolPerson, String, null: true

# crappy example because we have a diff cop for this method name but you get the idea
def coolPerson
  "It's you. You're the cool person."
end

# AFTER
field :cool_person, String, null: true
```

```stack trace
RuntimeError - Failed to implement Query.coolPerson, tried:

        - `Marketplace::Types::Query#cool_person`, which did not exist
        - `NilClass#cool_person`, which did not exist
        - Looking up hash key `:cool_person` or `"cool_person"` on ``, but it wasn't a Hash

        To implement this field, define one of the methods above (and check for typos)
```

config/default.yml
```diff
@@ -1,3 +1,10 @@
+Ezcater/GraphqlFieldsNaming:
+  EnforcedStyle: snake_case
+  Enabled: true
+  SupportedStyles:
+    - snake_case
+    - camelCase
+
```
tjwp: Do you want to restrict this cop to only apply to certain directories, e.g. `app/graphql`?

rusterholz:
This looks awesome. Excited to have it!

config/default.yml
tvec: Good call! just updated.

PR #91, opened 2020-03-06
add a cop for Timeout.timeout

halfwolf:
## What did we change?
- add a cop for ruby's `Timeout.timeout` 

## Why are we doing this?
- ruby's Timeout is bad. It caused stuck sidekiq workers that caused at least one crit sit. It's dangerous and cannot be trusted. 

## How was it tested?
- [x] Specs
- [ ] Locally


spec/rubocop/cop/ezcater/ruby_timeout_spec.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::Ezcater::RubyTimeout do
+  subject(:cop) { described_class.new }
+
+  let(:msgs) { [described_class::MSG] }
+
+  it "detects `Timeout.time`" do
```
tjwp: Typo: should be `Timeout.timeout`

tjwp:
Build was failing because it was picking up Simplecov > 0.18.0 which is incompatible with reporting test coverage to CodeClimate. 

Push a commit to restrict to < 0.18.0. See https://github.com/codeclimate/test-reporter/issues/413

aprescott:
I can fix up this PR and get it merged, since Colin is on vacation for the week.

PR #92, opened 2020-03-09
Detect Timeout.timeout in the expected cases

aprescott:
## What did we change?

As in the title.

## Why are we doing this?

`(send (const _ :Foo) :bar)` matches `Foo.bar` but not `Foo.bar(...)`.

## How was it tested?
- [x] Specs
- [x] Locally

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ This gem is moving onto its own [Semantic Versioning](https://semver.org/) schem
 
 Prior to v1.0.0 this gem was versioned based on the `MAJOR`.`MINOR` version of RuboCop. The first release of the ezcater_rubocop gem was `v0.49.0`.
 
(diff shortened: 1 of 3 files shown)


aprescott:
I've checked this against a repo to confirm it finds occurrences.

aprescott:
Grepping for `Timeout.timeout` counts:

```
$ ack -l -c 'Timeout.timeout' --ignore-dir={tmp,vendor,log,coverage,public,app/assets/javascripts}/ | cut -d ':' -f 2 | awk '{ sum += $1 } END { print sum }'
16
```

That count lines up with `rubocop --only Ezcater/RubyTimeout`, and the file list from `ack -l -c` lines up with the list from `rubocop --only Ezcater/RubyTimeout --format files`.

aprescott:
@tjwp indeed!

PR #93, opened 2020-04-23
Update rubocop dependencies

tjwp:
## What did we change?

Updated to the almost latest rubocop, rubocop-rspec and rubocop-rails (newly extracted).

## Why are we doing this?

I started looking at supporting Ruby 2.7 and found that needed to use a newer rubocop to support it.

I did not update to the very latest rubocop (v0.82) because it was released very recently and is already accumulating bug fixes.

I'm proposing this as a major version bump because it has been a while since we have updated these dependencies and some manual work (either regenerating TODO or fixing offenses) will be required when updating.

The rubocop project itself is also in transition -- attempting to reach 1.0 and support semver. From the config, you'll see that some cops have been renamed. New cops are now being introduced in a pending state. That is why you'll see them explicitly enabled here. This mechanism may change with a future release which is why I've separated them in the config.

## How was it tested?
- [x] Specs
- [ ] Locally


conf/rubocop.yml
```diff
@@ -139,3 +139,20 @@ Style/TrailingCommaInArrayLiteral:
 
 Style/TrailingCommaInHashLiteral:
   EnforcedStyleForMultiline: consistent_comma
+
+### New Cops
```
aprescott, line 143: To avoid the meaning of "New" here decaying over time, what do you think about something specifically around them being pending? "Cops pending in v0.81.0" might be too specific, but something along those lines?

tjwp: Modified the comment here with more info and broken by release.

PR #94, opened 2020-05-11
DEL-379 Adds cop for adding comment for migration best practices documentation to migration files

kdurr:
## [DEL-379](https://ezcater.atlassian.net/browse/DEL-379) What did we change?
- Added a cop to enforce having a comment with a link to the ezCater migration best practices doc.
- Tagged github's suggested peeps. Ignore if you have no opinions!
- Note: I have no idea if I did this correctly...

## Why are we doing this?
We had a crit-sit on the delivery squad a few months back that was caused by a mix up in migrations. We want to make sure that everyone follows the ezCater best practices for writing migrations and this will serve as a reminder with every new migration added.

## How was it tested?
- [x] Specs
- [ ] Locally

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ This gem is moving onto its own [Semantic Versioning](https://semver.org/) schem
 
 Prior to v1.0.0 this gem was versioned based on the `MAJOR`.`MINOR` version of RuboCop. The first release of the ezcater_rubocop gem was `v0.49.0`.
 
+## v2.1.0
+- Add `Ezcater/MigrationDocumentation` cop.
+
 ## v2.0.0
 - Update to `rubocop` v0.81.0, `rubocop-rspec` v1.38.1 and `rubocop-rails` v2.5.2.
 - This is being released as a major update because cops have been renamed so this is unlikely to be
--- a/lib/rubocop/cop/ezcater/migration_documentation.rb
+++ b/lib/rubocop/cop/ezcater/migration_documentation.rb
@@ -1,0 +1,67 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Ezcater
+      # Enforce use of migration best practices by adding comment of documentation url
+      #
+      # @example
+      #   # bad
+      #   # frozen_string_literal: true
+      #
+      #   class CreateUsers < ActiveRecord::Migration[5.1]
+      #     def change
+      #
+      #   # good
+      #   # frozen_string_literal: true
+      #   # https://ezcater.atlassian.net/wiki/spaces/POL/pages/53248012/Adding+a+migration
+      #
+      #   class CreateUsers < ActiveRecord::Migration[5.1]
+      #     def change
+      #
+      #
+      class MigrationDocumentation < Cop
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ This gem is moving onto its own [Semantic Versioning](https://semver.org/) schem
 
 Prior to v1.0.0 this gem was versioned based on the `MAJOR`.`MINOR` version of RuboCop. The first release of the ezcater_rubocop gem was `v0.49.0`.
 
+## v2.1.0
```
kdurr, line 9: Is this the correct version?

lib/rubocop/cop/ezcater/migration_documentation.rb
```diff
@@ -0,0 +1,67 @@
...
+      # @example
+      #   # bad
+      #   # frozen_string_literal: true
+      #
+      #   class CreateUsers < ActiveRecord::Migration[5.1]
+      #     def change
+      #
+      #   # good
+      #   # frozen_string_literal: true
+      #   # https://ezcater.atlassian.net/wiki/spaces/POL/pages/53248012/Adding+a+migration
```
zoso10, line 17: will this conflict with existing cops? i think we require a newline after the magic frozen string comment

lib/rubocop/cop/ezcater/migration_documentation.rb
```diff
@@ -0,0 +1,67 @@
...
+        MSG = "Missing the comment `#{MIGRATION_DOC_COMMENT}` " \
+              "at the top of your migration file."
+
+        def investigate(processed_source)
+          return if processed_source.tokens.empty?
+
+          ensure_comment(processed_source)
+        end
+
+        def autocorrect(_node)
```
zoso10, line 35: :raised_hands: for autocorrect

lib/rubocop/cop/ezcater/migration_documentation.rb
kdurr: ohhh good to know. well that's fun

kdurr: let me investigate and make changes if needed

kdurr:
>would this cause failures for existing migration files?

that would be most annoying so i hope not. i'll need to try to test it out

> i wonder if this is better suited in ezcater_migrations gem as part of the development checks 
urging the person to reference the doc then

that's a great idea. i have no strong opinions but having this there makes sense to me.

tvec:
> rubocop will ensure that folks are aware of that document, but i wonder if it's too late in the feedback loop of checks. i wonder if this is better suited in ezcater_migrations gem as part of the development checks urging the person to reference the doc then

👍 to this

 I'll often ignore comments (especially if they come from atlassian 😅) and what's nice about the migration gem is I believe it can give more targeted DB feedback to the user

CHANGELOG.md
kraila: if it requires manual changes to existing migration files, I would count that as not backwards compatible, and bump major version. But, if it leaves existing migrations as is, this looks correct.

kdurr:
Thank you all for your comments 🥇 
I'm going to close this and follow Tylers advice to have a message show up in the ezcater_migrations repo instead.

PR #95, opened 2020-09-30
Enable Rails/SaveBang

aprescott:
## What did we change?

Enabled [Rails/SaveBang](https://rubocop.readthedocs.io/projects/rails/en/stable/cops_rails/#railssavebang).

## Why are we doing this?

This rule catches problematic uses of non-`!` persistence methods like `save`:

```ruby
def foo
  x
  user.save # this might return false
  y
end
```

Without inspecting `user.save` here for `true`/`false`, we're more likely to introduce inconsistent/unexpected states that manifest later on.

`AllowImplicitReturn` seems too permissive if enabled, since it doesn't catch problematic `ary.each { x.save }` instances.

## How was it tested?
- [ ] Specs
- [x] Locally (via the same config in one of our repos)

### DIFF
--- a/conf/rubocop_rails.yml
+++ b/conf/rubocop_rails.yml
@@ -47,6 +47,11 @@ Rails/UnknownEnv:
     - staging
     - production
 
+Rails/SaveBang:
+  Enabled: true
+  AllowImplicitReturn: false
+  AutoCorrect: False
+
 Ezcater/RailsTopLevelSqlExecute:
   Description: 'Use `execute` instead of `ActiveRecord::Base.connection.execute` in migrations.'
   Enabled: true
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ This gem is moving onto its own [Semantic Versioning](https://semver.org/) schem
 
 Prior to v1.0.0 this gem was versioned based on the `MAJOR`.`MINOR` version of RuboCop. The first release of the ezcater_rubocop gem was `v0.49.0`.
 
+## Unreleased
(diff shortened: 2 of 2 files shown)


jaeger401 (approved):
I think the slight ugliness of having to say `return foo.save` is outweighed by the decreased risk of an unchecked true/false return value. I'd be interested to see how many violations we've got in ez-rails....

conf/rubocop_rails.yml
```diff
@@ -47,6 +47,10 @@ Rails/UnknownEnv:
     - staging
     - production
 
+Rails/SaveBang:
+  Enabled: true
+  AllowImplicitReturn: false
```
sagotsky, line 52: Do we want to disable autocorrection?  It's marked as unsafe.  It's not obvious to me what it autocorrects - does it replace methods with their bang version or does it save the result to a local var?

sagotsky (approved):

conf/rubocop_rails.yml
aprescott: It does a straight substitution from, e.g., `save` to `save!`. I'm +1 on disabling autocorrection, particularly since I think it helps enforce the understanding and habit.

jaeger401: According to the Slack thread, it apparently replaces with the bang equivalent.

kevinmook: I believe it autocorrects by replacing methods with their bang version. Disabling autocorrect seems like a good idea to me - let the developer explicitly choose the correct fix, else they may not have a chance to test/handle the error.

kevinmook (approved):

PR #96, opened 2021-04-09
Drop support for Ruby 2.5

tjwp:
## What did we change?

Changed earliest supported Ruby version to 2.6.

Set `TargetRubyVersion` to 2.6 in the `rubocop_gem` configuration used by ezCater gems.

## Why are we doing this?

Ruby 2.5 is now EOL.

## How was it tested?
- [x] Specs
- [ ] Locally


zoso10 (approved):

sagotsky (approved):
TY!

PR #98, opened 2021-05-18
Add cop for ensuring `EzFF.active?` is called with at least one of the required named arguments to go with our upcoming 2.0.0 release of the gem.

mikowitz:
## What did we change?

Add new cop `FeatureFlagActive` that ensures the new method signature for `EzcaterFeatureFlag.active?` is respected: https://github.com/ezcater/ezcater_feature_flag-client-ruby/pull/30

This cop will ensure that when `EzFF.active?` is called, it passes in the name of the feature flag to be evaluated, and at least one of either a `tracking_id` or a list of `identifiers` to ensure consistent evaluation.

## Why are we doing this?

We are upgrading the `EzFF` client API to make calling it more descriptive. Moving forward, the `active?` method should only be used when consistent evaluation against an identifier is desired, and the API now enforces passing one of those values. This cops will allow us to find any improper calls using the old API, and direct developers to upgrading either the arguments passed, or the method used.

## How was it tested?
- [x] Specs
- [ ] Locally

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ This gem is moving onto its own [Semantic Versioning](https://semver.org/) schem
 
 Prior to v1.0.0 this gem was versioned based on the `MAJOR`.`MINOR` version of RuboCop. The first release of the ezcater_rubocop gem was `v0.49.0`.
 
+## v2.3.0
+- Add `FeatureFlagActive` cop. This provides confidence that upgrading to `ezcater_feature_flag-client` v2.0.0, which
+    contains breaking API changes, can be done safely.
+
 ## v2.2.0
 - Require Ruby 2.6 or later.
 - Set `TargetRubyVersion` to 2.6 in `rubocop_gem` configuration.
--- a/lib/rubocop/cop/ezcater/feature_flag_active.rb
+++ b/lib/rubocop/cop/ezcater/feature_flag_active.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Ezcater
+      # Use `EzcaterFeatureFlag.active?` with a tracking_id or array of identifiers
+      #
+      # @example
+      #
+      #   # good
+      #   EzFF.active?("FlagName", tracking_id: "user:12345")
+      #   EzFF.active?("FlagName", identifiers: ["user:12345", "user:23456"])
+      #
+      #   # bad
+      #   EzFF.active?("FlagName")
+
+      class FeatureFlagActive < Cop
+
+        MSG = "`EzFF.active?` must be called with at least one of `tracking_id` or `identifiers`"
+
+        def_node_matcher :ezff_active_one_arg, <<-PATTERN
+          (send
+            (_ _ {:EzFF :EzcaterFeatureFlag}) :active? (str _))
+        PATTERN
+
+        def_node_matcher :args_matcher, <<-PATTERN
+          (send
+            (_ _ {:EzFF :EzcaterFeatureFlag}) :active?
+              (str _)
+              (_
+                (pair
+                  (sym {:tracking_id :identifiers})
+                  _)
+                ...))
+        PATTERN
+
+        def on_send(node)
+          if ezff_active_one_arg(node) || !args_matcher(node)
+            add_offense(node, location: :expression, message: MSG)
+          end
+        end
+      end
+    end
+  end
+end
--- a/spec/rubocop/cop/ezcater/feature_flag_active_spec.rb
+++ b/spec/rubocop/cop/ezcater/feature_flag_active_spec.rb
@@ -1,0 +1,55 @@
+# encoding utf-8
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::Ezcater::FeatureFlagActive, :config do
+  subject(:cop) { described_class.new(config) }
+  let(:flag_name) { "FeatureFlag #{rand(100)}" }
+  let(:tracking_id) { generate_tracking_id }
+
+  let(:msgs) { [described_class::MSG] }
+
+  before { inspect_source(line) }
+
+  %w(EzcaterFeatureFlag EzFF).each do |constant_name|
+    describe "calling #{constant_name}.active?" do
+      context "with a tracking_id" do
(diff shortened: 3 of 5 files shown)


spec/rubocop/cop/ezcater/feature_flag_active_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  it "errors when only one argument is passed" do
+    source = %[EzFF.active?("FlagName")]
+    inspect_source(source)
+    expect(cop.messages).to match_array(msgs)
+  end
+
+  it "works with the full module name" do
+    source = %[EzcaterFeatureFlag.active?("FlagName")]
```
sagotsky, line 34: Why does the full module behave differently?

spec/rubocop/cop/ezcater/feature_flag_active_spec.rb
```diff
@@ -0,0 +1,39 @@
+# encoding utf-8
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::Ezcater::FeatureFlagActive, :config do
+  subject(:cop) { described_class.new(config) }
+
+  let(:msgs) { [described_class::MSG] }
```
sagotsky, line 7: Should we be expecting anything about the contents of the message?  Someone could replace the constant definition with a poop emoji and this would still end up passing.

:8ball: because this is pretty nitpicky and may not be worth the trouble.

sagotsky (approved):
TY for this!

spec/rubocop/cop/ezcater/feature_flag_active_spec.rb
mikowitz: It doesn't, just wanted to ensure we're catching this behavior whether someone uses the full module name or the alias when they call the method.

sagotsky: Ugh, misread.  NM.

spec/rubocop/cop/ezcater/feature_flag_active_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    expect(cop.messages).to match_array(msgs)
+  end
+
+  it "errors when only one argument is passed" do
+    source = %[EzFF.active?("FlagName")]
+    inspect_source(source)
+    expect(cop.messages).to match_array(msgs)
+  end
+
+  it "works with the full module name" do
```
jmpage, line 33: This is not a change that absolutely needs to happen but it's good practice to explore a broader space of inputs:

Thoughts on using a bit of metaprogramming to ensure that we test each case for both `EzFF` and `EzcaterFeatureFlag`? You could also add in a bit of randomization to ensure that your test cases aren't tied to a particular implementation:

```
let(:flag_name) { generate(:feature_flag_name) }
let(:tracking_id) do
  # some code to generate a random id-like value
end
let(:message) { "..." }

before { inspect_source(line) }

%w(EzFF EzcaterFeatureFlag).each do |constant_name|
  describe "a line referring to the #{constant_name} constant" do
    context "and correctly uses a tracking_id" do
      let(:line) { "#{constant_name}.active?(#{flag_name}, tracking_id: #{tracking_id})" }
      
      it "does not report an issue" do
        expect(cop.messages).not_to include(msg)
      end
    end
    
    # ...
  end
end
```

jmpage (approved):

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ This gem is moving onto its own [Semantic Versioning](https://semver.org/) schem
 
 Prior to v1.0.0 this gem was versioned based on the `MAJOR`.`MINOR` version of RuboCop. The first release of the ezcater_rubocop gem was `v0.49.0`.
 
+## v2.3.0
+- Add `FeatureFlagActive` cop.
```
jmpage, line 10: 🎱  Thoughts on noting the reason behind this change to provide extra context to people who are upgrading? This could be especially helpful as someone knows that upgrading rubocop can provide extra confidence when subsequently upgrading ezcater ff.

jmpage (approved):

CHANGELOG.md
mikowitz: Done

jmpage (approved):

spec/rubocop/cop/ezcater/feature_flag_active_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+        it "reports an offense" do
+          expect(cop.messages).to match_array(msgs)
+        end
+      end
+    end
+  end
+
+  def generate_tracking_id
+    [
+      %w(user brand caterer).sample,
```
djcp, line 51: :8ball: In this case I think we're safe, but I have found that randomness in tests can sometimes lead to flakiness.

lib/rubocop/cop/ezcater/feature_flag_active.rb
```diff
@@ -0,0 +1,45 @@
...
+      class FeatureFlagActive < Cop
+
+        MSG = "`EzFF.active?` must be called with at least one of `tracking_id` or `identifiers`"
+
+        def_node_matcher :ezff_active_one_arg, <<-PATTERN
+          (send
+            (_ _ {:EzFF :EzcaterFeatureFlag}) :active? (str _))
+        PATTERN
+
+        def_node_matcher :args_matcher, <<-PATTERN
```
djcp, line 26: Having never written a cop, this syntax is both confusing but makes sense on further inspection. Neat!

djcp (approved):

spec/rubocop/cop/ezcater/feature_flag_active_spec.rb
sagotsky: +1.  If all cases are worth testing I'd rather test all of them every time.  It might be worth embedding the randomized piece in the error output for reproducibility's sake

lib/rubocop/cop/ezcater/feature_flag_active.rb
mikowitz: Yep! I felt the same way! There's a nifty command line tool `ruby-parse` that was very helpful:

```shell
~$ ruby-parse -e 'EzFF.active?("foo", bar: "baz")'
(send
  (const nil :EzFF) :active?
  (str "foo")
  (kwargs
    (pair
      (sym :bar)
      (str "baz"))))
```

PR #99, opened 2021-05-19
Add fix for FeatureFlagActive cop

mikowitz:
## What did we change?

The previous logic was overly broad in what it would flag. Anything that parsed to use `send` in the AST but wasn't a call to `EzFF.active?` would be flagged an invalid. This includes things like `require` and `.config do` blocks.

## Why are we doing this?

The previous logic was catching things that shouldn't have been caught

## How was it tested?
- [x] Specs
- [ ] Locally


sagotsky (approved):
Weird bug.  Nice turnaround time though.

djcp (approved):
Nice catch!

PR #100, opened 2021-05-28
Fix `FeatureFlagActive` cop so that it allows feature flag names to be variables in addition to string

mikowitz:
## What did we change?

Fix `FeatureFlagActive` cop so that it allows feature flag names to be variables in addition to string

whereas previously only
```
EzFF.active?("FeatureFlag", tracking_id: ...)
```

was considered valid, we now also allow
```
EzFF.active?(defined_feature_flag_name_var, ...)
EzFF.active?(@feature_flag_name_ivar, ...)
```

## Why are we doing this?

In its previous state, this cop expected the feature flag name to be passed in as a string, and was erroneously erroring when the name was passed in as a variable. This was something we forgot to take into account in the first iteration of this, but is a reasonable and expected use case, so we're allowing for it now.

## How was it tested?
- [x] Specs
- [ ] Locally

### DIFF
--- a/spec/rubocop/cop/ezcater/feature_flag_active_spec.rb
+++ b/spec/rubocop/cop/ezcater/feature_flag_active_spec.rb
@@ -8,16 +8,45 @@ RSpec.describe RuboCop::Cop::Ezcater::FeatureFlagActive, :config do
   let(:tracking_id) { generate_tracking_id }
 
   let(:msgs) { [described_class::MSG] }
+  let(:first_params_msgs) { [described_class::FIRST_PARAM_MSG] }
 
   before { inspect_source(line) }
 
-  %w(EzcaterFeatureFlag EzFF).each do |constant_name|
+  %w(EzcaterFeatureFlag EzFF ::EzFF ::EzcaterFeatureFlag).each do |constant_name|
     describe "calling #{constant_name}.active?" do
       context "with a tracking_id" do
         let(:line) { %[#{constant_name}.active?("#{flag_name}", tracking_id: "#{tracking_id}")] }
 
         it "does not report an offense" do
           expect(cop.offenses).to be_empty
+        end
+      end
+
+      context "with a variable instead of a flag name" do
+        let(:line) do
+          %[def evaluate(my_flag_name_var)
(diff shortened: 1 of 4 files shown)


jgoggles (approved):
Thanks!

sagotsky (approved):

jaeger401 (approved):

spec/rubocop/cop/ezcater/feature_flag_active_spec.rb
```diff
@@ -11,7 +11,7 @@
 
   before { inspect_source(line) }
 
-  %w(EzcaterFeatureFlag EzFF).each do |constant_name|
+  %w(EzcaterFeatureFlag EzFF ::EzFF).each do |constant_name|
```
jaeger401, line 14: 🎱 Super unlikely that we'd need it, but is there value in being pedantic and adding `::EzcaterFeatureFlag` too while we're adding scoped options here?

mikowitz: Probably. In all honestly neither really needs to be here. This is a relic from when I thought the namespacing was the issue, but those specs passed without needing to change anything, as evidenced by the eventual change I figured out. But I think there's some value in having these here as regression-proofing if Ruby AST shapes ever change.

spec/rubocop/cop/ezcater/feature_flag_active_spec.rb
```diff
@@ -11,7 +11,7 @@
 
   before { inspect_source(line) }
 
-  %w(EzcaterFeatureFlag EzFF).each do |constant_name|
+  %w(EzcaterFeatureFlag EzFF ::EzFF ::EzcaterFeatureFlag).each do |constant_name|
```
djcp, line 15: It doesn't look like we needed to do anything to fix namespacing in this cop, but I appreciate adding additional tests when they're ones like this that probably execute very quickly.

djcp (approved):
Looking good! I assume you checked any other cops we've touched recently as well.

jaeger401 (approved):
:shipit:

PR #102, opened 2021-06-04
Bump to latest rubocop 1.16.0

### LINKED ISSUE #101: Upgrade to newer rubocop
@jaeger401 
> With new cops in a pending state by default from 1.0 onward, bumping Rubocop should get a whole heck of a lot easier. We may have one last big heavy lift to get from here to 1.0 though.

sagotsky:
Fixes #101 

## What did we change?

- Bumped rubocop to most recent version.
- Bumped rubocop-{rails,rspec} as well
- Disabled a large number of new cops https://github.com/ezcater/ezcater_rubocop/pull/102/commits/872b1634d17e34afc796a326234e8a4b3395586e
- Marked this as a new major version

## Why are we doing this?

1. Application fitness 🏋️ 
2. After 1.0, new cops will be introduced in a default state.  This will make future version bumps much easier.
3. Disabling many of those cops is likely going to be my spicy take of the day.  Bluntly speaking, we should have been doing more minor version bumps along the way.  That would have introduced those cops in pending states and then phased them in to an enabled state.  I suppose we could retroactively step through all the rubocop version bumps, but I'd rather just fast forward past that step.  
4. We've upgraded rubocop major versions.  They had breaking changes.  Therefore we inherited their breaking changes.

## How was it tested?
- [x] Specs
- [ ] Locally

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,31 @@ This gem is moving onto its own [Semantic Versioning](https://semver.org/) schem
 
 Prior to v1.0.0 this gem was versioned based on the `MAJOR`.`MINOR` version of RuboCop. The first release of the ezcater_rubocop gem was `v0.49.0`.
 
+## v3.0.0
+- Upgrade rubocop:       1.16.0
+- Upgrade rubocop-rails: 2.10.1
+- Upgrade rubocop-rspec: 2.3.0
+- This is a major upgrade because a large number of cops have been introduced, tweaked, or renamed.  Here is an aggregate of the breaking changes introduced
+  - RuboCop assumes that Cop classes do not define new `on_<type>` methods at runtime (e.g. via `extend` in `initialize`).
+  - Enable all pending cops for RuboCop 1.0.
+  - Change logic for cop department name computation. Cops inside deep namespaces (5 or more levels deep) now belong to departments with names that are calculated by joining module names starting from the third one with slashes as separators. For example, cop `Rubocop::Cop::Foo::Bar::Baz` now belongs to `Foo/Bar` department (previously it was `Bar`).
+  - `RegexpNode#parsed_tree` now processes regexps including interpolation (by blanking the interpolation before parsing, rather than skipping).
+  - Cop `Metrics/AbcSize` now counts ||=, &&=, multiple assignments, for, yield, iterating blocks. `&.` now count as conditions too (unless repeated on the same variable). Default bumped from 15 to 17. Consider using `rubocop -a --disable-uncorrectable` to ease transition.
+  - Cop `Metrics/PerceivedComplexity` now counts `else` in `case` statements, `&.`, `||=`, `&&=` and blocks known to iterate. Default bumped from 7 to 8. Consider using `rubocop -a --disable-uncorrectable` to ease transition.
+  - Extensive refactoring of internal classes `Team`, `Commissioner`, `Corrector`. `Cop::Cop#corrections` not completely compatible. See Upgrade Notes.
+  - `rubocop -a / --autocorrect` no longer run unsafe corrections; `rubocop -A / --autocorrect-all` run both safe and unsafe corrections. Options `--safe-autocorrect` is deprecated.
+  - Order for gems names now disregards underscores and dashes unless `ConsiderPunctuation` setting is set to `true`.
+  - Cop `Metrics/CyclomaticComplexity` now counts `&.`, `||=`, `&&=` and blocks known to iterate. Default bumped from 6 to 7. Consider using `rubocop -a --disable-uncorrectable` to ease transition.
+  - Remove support for unindent/active_support/powerpack from `Layout/HeredocIndentation`, so it only recommends using squiggy heredoc.
+  - Change the max line length of `Layout/LineLength` to 120 by default.
+  - Inspect all files given on command line unless `--only-recognized-file-types` is given.
+  - Enabling a cop overrides disabling its department.
+  - Renamed `Layout/Tab` cop to `Layout/IndentationStyle`.
+  - Drop support for Ruby 2.3.
+  - Drop support for ruby 2.4.
+  - Retire `RSpec/InvalidPredicateMatcher` cop.
+  - Enabled pending cop (`RSpec/StubbedMock`).
+
 ## v2.4.0
 - Fix `FeatureFlagActive` cop so that it allows feature flag names to be variables in addition to strings.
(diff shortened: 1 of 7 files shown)


sagotsky:
Looks like there's no CI for this repo.  I get failures on the following specs locally.  

`rspec ./spec/rubocop/cop/ezcater/graphql_fields_naming_spec.rb` 
`rspec ./spec/rubocop/cop/ezcater/style_dig_spec.rb`

jaeger401 (approved):
Gotta love a +3/-3 that adds this much value. Thanks for taking this on!

jaeger401:
Probably worth updating the PR description to "Fixes #101" so merging this closes the issue automagically?

sagotsky:
Good call!  I forgot how much I miss that kind of github magic.

jmpage (commented):
Thoughts on testing this a prerelease of this version against a repo to get an idea of what might break with this upgrade?

Thoughts on updating CHANGELOG.md and the VERSION, as appropriate?

sagotsky:
@jmpage Yes, to all.

sagotsky:
The prerelease testing is what blocked a new version number.  I'm not sure just how breaking this is yet.

conf/rubocop.yml
```diff
@@ -146,6 +153,181 @@ Style/TrailingCommaInHashLiteral:
...
+  Enabled: false
+
+Style/SingleArgumentDig:
+  Enabled: false
+
+Style/SlicingWithRange:
+  Enabled: false
+
+Style/SoleNestedConditional:
+  Enabled: false
```
jmpage, line 329: Thoughts on opening an issue on this project for evaluating these for inclusion and include a checkbox TODO list in the description of each disabled cop?

spec/rubocop/cop/ezcater/style_dig_spec.rb
```diff
@@ -4,7 +4,7 @@
 RSpec.describe RuboCop::Cop::Ezcater::StyleDig, :config do
   subject(:cop) { described_class.new(config) }
 
-  let(:ruby_version) { 2.3 }
+  let(:ruby_version) { 2.6 }
```
jmpage, line 7: What prompted this change?

CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ This gem is moving onto its own [Semantic Versioning](https://semver.org/) schem
 
 Prior to v1.0.0 this gem was versioned based on the `MAJOR`.`MINOR` version of RuboCop. The first release of the ezcater_rubocop gem was `v0.49.0`.
 
+## v3.0.0
+- Upgrade rubocop:       1.16.0
+- Upgrade rubocop-rails: 2.10.1
+- Upgrade rubocop-rspec: 2.3.0
+- This is a major upgrade because a large number of cops have been introduced, tweaked, or renamed.
```
jmpage, line 13: Thoughts on including (or linking to) a list of renamed cops so that people can easily update their `rubocop:disable` statements?

CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ This gem is moving onto its own [Semantic Versioning](https://semver.org/) schem
 
 Prior to v1.0.0 this gem was versioned based on the `MAJOR`.`MINOR` version of RuboCop. The first release of the ezcater_rubocop gem was `v0.49.0`.
 
+## v3.0.0
+- Upgrade rubocop:       1.16.0
+- Upgrade rubocop-rails: 2.10.1
+- Upgrade rubocop-rspec: 2.3.0
```
jmpage, line 12: Are there any official upgrade guides that we can link to to make bumps easier for folks?

jmpage (commented):
Thoughts on testing this on a non-ez-rails app as well, just to see how disruptive it might be?

CHANGELOG.md
sagotsky: Probably.  I haven't gone digging yet.

CHANGELOG.md
sagotsky: Sure.  Is the changelog the appropriate home for that?

spec/rubocop/cop/ezcater/style_dig_spec.rb
sagotsky: Not gonna lie, I couldn't completely figure this one out.  At least not without spending more time on it.

What I think is going on is that this cop had a minimum ruby version that it applied to.  The test was making assertions about the style given a particular version.  `ruby_version` isn't mentioned anywhere else in the file, so I assume it's provided by rubocop, but that's where I stopped digging in.  

I figured that since the gem required 2.6 or higher, we'd never be looking at a ruby lower than that.  Bumping the version here fixed the spec.

sagotsky:
@jmpage I'd be game for testing on other repos.  Do you have any in mind?  If not I'll just run this against whatever is checked out locally.

CHANGELOG.md
jmpage: It's the thing that dependabot will excerpt in a bump PR and the first place someone should look when determining changes. If the list is short, it's probably fine to include it. If it's longer, it may make sense to create a separate file in a `docs/migrations/` folder and link to that, or link to an external resource that lives in the rubocop repos.

conf/rubocop.yml
jmpage: I figure that GH issues let's us punt the issue of figuring out what to turn on from this PR while leaving us with a reminder to get back to it (which probably requires some deliberation for each cop).

sagotsky: https://github.com/ezcater/ezcater_rubocop/issues/104

I think it's a good home for repos that don't have jira projects.

CHANGELOG.md
sagotsky: I've rolled up the breaking changes from the three dependencies into our changelog.

jmpage (approved):

jmpage:
@sagotsky authentication-rails, identity-rails or ezcater-email-service are reasonably "simple" repos that you could test it against.

sagotsky:
Tested against a handful of repos.  All had single digit offenses.  `RequiredRubyVersion` needed to be fixed everywhere (it was the only offense for coupons-rails, ezcater-identity, and authentication-rails`.  club-soda-rails was the biggest repo and that weight in at 8 offenses, all of which were redundant.  I'm comfortable releasing like this.

jmpage:
🎉 :shipit: 

Thank you @sagotsky!

PR #105, opened 2022-03-18
Fix the RspecDotNotSelfDot cop to match offenses

BoboFraggins:
## What did we change?

Fix broken rspec_dot_not_self_dot cop.
Make this gem pass its own rubocop.

## Why are we doing this?

I was experimenting with Rails 7 and Ruby 3.1 and needed to update the underlying gems and that uncovered a cop that didn't work.

## How was it tested?
- [x] Specs
- [x] Locally


ezcater_rubocop.gemspec
```diff
@@ -42,17 +42,17 @@ Gem::Specification.new do |spec|
...
+  spec.add_development_dependency "rake", "~> 13.0"
+  spec.add_development_dependency "rspec", "~> 3.11"
   spec.add_development_dependency "rspec_junit_formatter"
   spec.add_development_dependency "simplecov", "< 0.18.0"
 
-  spec.add_runtime_dependency "parser", "!= 2.5.1.1"
-  spec.add_runtime_dependency "rubocop", "~> 1.16.0"
-  spec.add_runtime_dependency "rubocop-rails", "~> 2.10.1"
-  spec.add_runtime_dependency "rubocop-rspec", "~> 2.3.0"
+  spec.add_runtime_dependency "parser", "~> 3.1"
```
jmpage, line 54: This new constraint seems less permissive than the previous one. What is the requirement that pushed us to update this?

lib/rubocop/cop/ezcater/graphql_fields_naming.rb
```diff
@@ -66,7 +66,7 @@ def method_name(node)
         end
 
         def message(style)
-          format(MSG, style: style)
+          format(MSG, style:)
```
jmpage, line 69: Will this change be backwards compatible with older versions of ruby that we still support (such as 2.7)?

jmpage (changes requested):
Would you mind adding an appropriate entry or set of entries in `CHANGELOG.md` so that people know what changed when upgrading?

.rubocop.yml
```diff
@@ -1,5 +1,6 @@
 AllCops:
-  TargetRubyVersion: 2.6
+  TargetRubyVersion: 3.1
```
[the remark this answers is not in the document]
BoboFraggins: I believe this file is specifying which ruby version to test the gem's code against, not which version the client app uses. I will double check on that, though.

ezcater_rubocop.gemspec
BoboFraggins: Sure, I was playing with updated gems and ruby when I hit the problems. I can move those to `>=` of the previous versions.

.rubocop.yml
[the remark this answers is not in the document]
BoboFraggins: There's a rubocop forcing this and the version in the gemspec to match, so I reverted this.

ezcater_rubocop.gemspec
BoboFraggins: I made them more permissive.

lib/rubocop/cop/ezcater/graphql_fields_naming.rb
BoboFraggins: Yes, indeed. I reverted and added a Rubocop disable to get rid of the warning.

ezcater_rubocop.gemspec
jmpage: My concern is mostly about supporting a reasonable range so that other applications do not have to upgrade these gems when upgrading `ezcater_rubocop` unless ezcater rubocop is explicitly incompatible with the older versions.

BoboFraggins: I agree. I set the dependencies to be >= the previous version. I'm looking for an app that uses the previous version and testing against that, but every app that I have worked on caps the version of this gem to really old versions. I will keep looking for a candidate rails app to test against.

ezcater_rubocop.gemspec
```diff
@@ -42,17 +42,17 @@ Gem::Specification.new do |spec|
...
+  spec.add_runtime_dependency "rubocop", "~> 1.26.0"
+  spec.add_runtime_dependency "rubocop-rails", "~> 2.13.2"
+  spec.add_runtime_dependency "rubocop-rspec", "~> 2.9.0"
```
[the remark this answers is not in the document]
BoboFraggins: Definitely. I moved the minimum versions back to what they were set to on last version of the gem.

BoboFraggins:
Changelog entry added.

lib/rubocop/cop/ezcater/graphql_fields_naming.rb
```diff
@@ -66,7 +66,7 @@ def method_name(node)
         end
 
         def message(style)
-          format(MSG, style: style)
+          format(MSG, style: style) # Rubocop:disable Style/HashSyntax
```
jaeger401, line 69: I pulled this branch down and yanked the `disable` out and Rubocop didn't bark. Could we remove this or is there something else here that I missed?

lib/ezcater_rubocop/version.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module EzcaterRubocop
-  VERSION = "3.0.0"
+  VERSION = "3.1.0"
```
jaeger401, line 4: If we do decide that unpinning the gem versions is OK after all, I think it would warrant a major version bump.

lib/rubocop/cop/ezcater/graphql_fields_naming.rb
BoboFraggins: If you try with the newest rubocop, it barks on that, so to support newer dependencies, it should probably stay.

lib/ezcater_rubocop/version.rb
BoboFraggins: Sure, makes sense.

lib/rubocop/cop/ezcater/rspec_dot_not_self_dot.rb
```diff
@@ -25,26 +25,39 @@ module Ezcater
...
-        def on_send(node)
-          example_group?(node) do |doc|
-            if doc.source.match?(SELF_DOT_REGEXP)
-              add_offense(doc, location: :expression, message: SELF_DOT_MSG)
-            elsif doc.source.match?(COLON_COLON_REGEXP)
-              add_offense(doc, location: :expression, message: COLON_COLON_MSG)
-            end
+        def_node_matcher :example_group?, <<~PATTERN
+          (block
+            (send #rspec? {#{EXAMPLE_GROUP_IDENTIFIERS.map { |i| ":#{i}" }.join(' ')}}
```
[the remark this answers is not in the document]
BoboFraggins: That was removed in this commit:

https://github.com/rubocop/rubocop-rspec/commit/e515d215e82b5abd12b64134dbf8728b5431002e

That is the root cause of why the matcher doesn't work in the main branch of this gem and why the tests error out when run.

spec/rubocop/cop/ezcater/rspec_dot_not_self_dot_spec.rb
```diff
@@ -45,9 +45,21 @@
...
+
+    it "accepts #{name} with `::` not at the start of the string" do
+      source = "#{name} \"something ::class_method\" do\nend"
+      inspect_source(source)
+      expect(cop.offenses).to be_empty
+    end
   end
 
-  RuboCop::RSpec::Language::Examples::ALL.each do |name|
+  described_class::EXAMPLE_IDENTIFIERS.each do |name|
```
jaeger401, line 62: 🎱  Similar to above, what's the rationale for changing this? Is there a reason to ignore `skip`, `pending`, and the `x`/`f`-prefix shortcuts?

ezcater_rubocop.gemspec
```diff
@@ -44,15 +44,15 @@ Gem::Specification.new do |spec|
...
+  spec.add_development_dependency "simplecov"
 
-  spec.add_runtime_dependency "parser", "!= 2.5.1.1"
-  spec.add_runtime_dependency "rubocop", "~> 1.16.0"
-  spec.add_runtime_dependency "rubocop-rails", "~> 2.10.1"
-  spec.add_runtime_dependency "rubocop-rspec", "~> 2.3.0"
+  spec.add_runtime_dependency "parser", ">= 2.5"
+  spec.add_runtime_dependency "rubocop", ">= 1.16.0"
+  spec.add_runtime_dependency "rubocop-rails", ">= 2.10.1"
+  spec.add_runtime_dependency "rubocop-rspec", ">=  2.3.0"
```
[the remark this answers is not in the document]
sagotsky: > @sagotsky kindly kicked off last year 

I can't believe that task has taken so long to resolve :'(  And by resolve, I mean sit in my inbox forever.

lib/rubocop/cop/ezcater/rspec_dot_not_self_dot.rb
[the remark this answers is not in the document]
jaeger401: Aha, gotcha. Perhaps it's worth setting to the last known good value then?

ezcater_rubocop.gemspec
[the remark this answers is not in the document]
BoboFraggins: There seems to be a difference of opinions on dependency versioning. Would it make sense for me to split this into two PRs?

1. Fix the rspec_dot_not_self_dot.rb spec and do a patch update.
2. Update to latest dependencies as a major update?

Also, I see this still uses "master", but most other repos have switched to "main". Should I fix that, too?

lib/rubocop/cop/ezcater/rspec_dot_not_self_dot.rb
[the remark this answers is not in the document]
BoboFraggins: I will incorporate the 'x' and 'f' versions into the lists.

BoboFraggins:
I have backed out all of the dependency changes, and now it is just fixing the broken cop and making tests/rubocop pass on this repo as a patch update.

sagotsky (approved):

jaeger401 (approved):
Thanks for being flexible on this one!

jmpage (approved):

PR #107, opened 2022-03-23
Update to latest ruby and dependency gems

BoboFraggins:
## What did we change?

Update to latest ruby and dependency gems

## Why are we doing this?

Greenfield project should probably not start with the tech debt of using older dependencies. The 3.x version line of this gem will be for legacy apps and the 4.x will be for Ruby 3.1 & Rails 7 apps.

## How was it tested?
- [x] Specs
- [x] Locally


CHANGELOG.md
```diff
@@ -6,7 +6,12 @@ This gem is moving onto its own [Semantic Versioning](https://semver.org/) schem
 
 Prior to v1.0.0 this gem was versioned based on the `MAJOR`.`MINOR` version of RuboCop. The first release of the ezcater_rubocop gem was `v0.49.0`.
 
+## v4.0.0
+
+- Update to Ruby 3.1 and latest dependency gems
```
jmpage, line 11: Is this a breaking change, as implied by the major version bump? If so, is there a way that we could support ruby 2.7 and 3.1 simultaneously?

.tool-versions
```diff
@@ -1 +1 @@
-ruby 2.7.3
```
jmpage, line 1: [CI] is still configured to run tests against ruby 2.7

[CI]: https://github.com/ezcater/ezcater_rubocop/blob/49f36ff06d6ac9dd2fdc539bfb519c82fc7f374f/.circleci/config.yml#L7

ezcater_rubocop.gemspec
```diff
@@ -44,15 +44,15 @@ Gem::Specification.new do |spec|
...
+  spec.add_development_dependency "rspec", "~> 3.11"
   spec.add_development_dependency "rspec_junit_formatter"
-  spec.add_development_dependency "simplecov", "< 0.18.0"
+  spec.add_development_dependency "simplecov"
 
-  spec.add_runtime_dependency "parser", "!= 2.5.1.1"
-  spec.add_runtime_dependency "rubocop", "~> 1.16.0"
-  spec.add_runtime_dependency "rubocop-rails", "~> 2.10.1"
-  spec.add_runtime_dependency "rubocop-rspec", "~> 2.3.0"
+  spec.add_runtime_dependency "parser", "~> 3.1"
```
jmpage, line 54: The new version constraint is actually more constrained than the past one, which still allowed project to use parser >= 3.1. Is there a particular reason why ezcater_rubocop can only support ~> 3.1 going forward?

ezcater_rubocop.gemspec
```diff
@@ -44,15 +44,15 @@ Gem::Specification.new do |spec|
...
+  spec.add_runtime_dependency "rubocop", "~> 1.26"
+  spec.add_runtime_dependency "rubocop-rails", "~> 2.14"
+  spec.add_runtime_dependency "rubocop-rspec", "~> 2.9"
```
jmpage, line 57: Thoughts on incrementally adopting these changes in 3.x so that apps dependent on ezcater_rubocop can benefit from them?

jmpage (changes requested):

.tool-versions
BoboFraggins: CircleCI is not being invoked on pushes. I assume that is how the specs stopped working and wasn't noticed. Should this be moved to building via Jenkins, or do we still have a CircleCI subscription? I'm happy to do either.

CHANGELOG.md
BoboFraggins: No, I must have misunderstood what was being asked. Let me try a new PR here.

.tool-versions
jmpage: 😬  it looks like it's running, but it's not reporting the status back for some reason.

https://app.circleci.com/pipelines/github/ezcater/ezcater_rubocop

ezcater_rubocop.gemspec
BoboFraggins: Can you re-review it with my new dependency specifications?

lib/rubocop/cop/ezcater/rspec_dot_not_self_dot.rb
```diff
@@ -35,8 +35,8 @@ class RspecDotNotSelfDot < Cop
           %w(it specify example scenario).map { |identifier| "#{prefix}#{identifier}" }
         end.flatten + %w(its focus skip)).freeze
 
-        SELF_DOT_REGEXP = /\Aself\./.freeze
-        COLON_COLON_REGEXP = /\A(\:\:)/.freeze
+        SELF_DOT_REGEXP = /\Aself\./
+        COLON_COLON_REGEXP = /\A(\:\:)/
```
jmpage, line 39: Should we continue to explicitly freeze these for ruby 2.7 support?

jmpage (approved):

lib/rubocop/cop/ezcater/rspec_dot_not_self_dot.rb
BoboFraggins: Sure, I can do that.

PR #108, opened 2022-04-18
Add rubocop-graphql helpers and release 4.0.0

drapergeek:
## What did we change?
This adds the [rubocop-graphql](https://github.com/DmitryTsepelev/rubocop-graphql) helpers. We've been using these in ez-rails for a while now and they've been useful. This also removes the custom cop `Ezcater/GraphqlFieldsNaming` as it's covered by a cop in the new gem.

## How was it tested?
- [ ] Specs
- [X] Locally


config/default.yml
```diff
@@ -59,3 +50,15 @@ Ezcater/StyleDig:
...
   Description: "Disallow use of `Timeout.timeout` because it is unsafe and can cause unexpected behavior."
   Enabled: true
+
+GraphQL/FieldDescription:
+  Enabled: false
+GraphQL/ArgumentDescription:
+  Enabled: false
+GraphQL/ObjectDescription:
+  Include:
+    - "**/types/**/*.rb"
```
drapergeek, line 60: I left this wide open because I don't think we want to define that the structure is `app/graphql/**`. We're using packwerk in marketplace hatchery so the directory looks more like `app/packages/<packagename>/graphql`.

config/default.yml
```diff
@@ -59,3 +50,15 @@ Ezcater/StyleDig:
 Ezcater/RubyTimeout:
   Description: "Disallow use of `Timeout.timeout` because it is unsafe and can cause unexpected behavior."
   Enabled: true
+
+GraphQL/FieldDescription:
+  Enabled: false
+GraphQL/ArgumentDescription:
+  Enabled: false
```
jmpage, line 57: Tangential: I would love for us to adopt these at some point.
It would be great if we could also adopt the no-self-references rule from the menus team for these as well when/if we adopt these in the future.

lib/ezcater_rubocop/version.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module EzcaterRubocop
-  VERSION = "3.0.2"
+  VERSION = "3.0.3"
```
jmpage, line 4: Per [semver](https://semver.org/), should this be a minor bump given that it's adding a new feature?

jmpage (changes requested):

config/default.yml
drapergeek: No day like today 🤷 Given that this is already a breaking change, this may be the best time to roll it out if you're so inclined.

lib/ezcater_rubocop/version.rb
jmpage: I'd be okay with a major version bump.

jmpage (approved):

PR #110, opened 2022-04-19
Add support for testing multiple rubies

drapergeek:
## What did we change?
* Update to the 2.1 version of CircleCI config
* Add support for testing multiple rubies. We require 2.6 or greater but we're only testing on 3.1 currently. This helps cover other ruby versions
* Remove the bundle caching - we don't have a gemfile lock so it's not as reliable to rebuild the same build each time. It's worth doing a fresh pull (the entire CI run only takes 30 seconds)

## How was it tested?
- [X] Specs
- [X] Locally

### DIFF
--- a/.circleci/working_config.yaml
+++ b/.circleci/working_config.yaml
@@ -1,0 +1,78 @@
+version: 2.1 # Use 2.1 to enable using orbs and other features.
+
+# Declare the orbs that we'll use in our config.
+# read more about orbs: https://circleci.com/docs/2.0/orb-intro/
+orbs:
+  ruby: circleci/ruby@1.0
+  node: circleci/node@2
+
+jobs:
+  build: # our first job, named "build"
+    docker:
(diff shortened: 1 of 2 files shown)


.circleci/working_config.yaml
```diff
@@ -0,0 +1,78 @@
+version: 2.1 # Use 2.1 to enable using orbs and other features.
```
jmpage, line 1: What's the purpose of this config? It appears to be loading database schema.

drapergeek: Whoops 😮 I need to kill this file. I was using this as a working reference while fixing all the broken schema changes from 2 to 2.1

ransombriggs (approved):

ransombriggs:
I am fine with losing the caching, when it breaks it leads to really confusing errors and at my previous job we had to keep a version number on it so we could manually bust it in a PR, it was annoying.

jmpage (approved):

PR #111, opened 2022-06-16
Add a cop for enforcing Graphql type layout

drapergeek:
## What did we change?
This adds a new cop for enforcing the order of Graphql type files.

## Why are we doing this?
We outlined our preferred order [here](https://github.com/ezcater/developer-handbook/pull/160) and now we want to enforce it using rubocop.

## How was it tested?
- [X] Specs
- [x] Locally

### DIFF
--- a/lib/rubocop/cop/ezcater/graph_ql_type_file_order.rb
+++ b/lib/rubocop/cop/ezcater/graph_ql_type_file_order.rb
@@ -77,7 +77,7 @@ module RuboCop
           end
 
           def resolve_method?
-            name.start_with?("resolve_")
+            name == "resolve"
           end
 
           def dsl_method?
--- a/spec/rubocop/cop/ezcater/graph_ql_type_file_order_spec.rb
+++ b/spec/rubocop/cop/ezcater/graph_ql_type_file_order_spec.rb
@@ -53,6 +53,25 @@ RSpec.describe RuboCop::Cop::Ezcater::GraphQlTypeFileOrder do
     expect(cop.messages).to match_array(["load_recipe should be positioned after load_cart"])
   end
 
+  it "enforces the order for the resolve" do
+    source = <<~SOURCE
+      class MyType
+        description "A great type."
+
+        def resolve(cart:)
+          Resolver.call(cart)
+        end
+
+        def load_cart(id)
+          CartLoader.call(id)
+        end
+      end
+    SOURCE
+
+    inspect_source(source)
+    expect(cop.messages).to match_array(["resolve should be positioned after load_cart"])
+  end
+
   it "handles unknown methods mixed between DSL calls" do
     source = <<~SOURCE
       class MyType


lib/rubocop/cop/ezcater/graph_ql_type_file_order.rb
```diff
@@ -0,0 +1,112 @@
...
+          end
+
+          if is_gql_type
+            gql_offenses.each do |offense|
+              add_offense(offense[:node], location: :expression, message: offense[:message])
+            end
+          end
+        end
+
+        private # rubocop:disable Lint/UselessAccessModifier
```
drapergeek, line 48: The rubocop matcher doesn't seem to see the `def_node_search` or the class definition as being "useful". When I added a method here just to test, it stopped complaining.

lib/rubocop/cop/ezcater/graph_ql_type_file_order.rb
```diff
@@ -0,0 +1,112 @@
...
+          def load_method?
+            name.start_with?("load_")
+          end
+
+          def resolve_method?
+            name == "resolve"
+          end
+
+          def dsl_method?
+            !(ORDERED_DSL_METHODS - %w(extend include)).index(name).nil?
```
jaeger401, line 84: 🎱 Am I missing a nuance or could we use `include?` here instead?

lib/rubocop/cop/ezcater/graph_ql_type_file_order.rb
```diff
@@ -0,0 +1,114 @@
...
+              is_gql_type = true
+            end
+
+            next unless previous > current
+
+            msg = "#{previous.name} should be positioned after #{current.name}"
+            gql_offenses << { node: previous.node, message: msg }
+          end
+
+          if is_gql_type
```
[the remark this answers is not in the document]
jaeger401: My $0.02: I think I'd be fine with trying this by keying off the method names -- they seem unusual enough at the class definition level that I'd expect them to be isolated to GQL files. If we see reports of false positives, then we can take action to address it at that point with the understanding that the affected dev can always just disable the cop in the interim.

lib/rubocop/cop/ezcater/graph_ql_type_file_order.rb
```diff
@@ -0,0 +1,112 @@
...
+        ORDERED_DSL_METHODS = %w(
+          extend
+          include
+          description
+          pundit_policy_class
+          pundit_role
+          argument
+          field
+          user_errors_field
+        ).freeze
```
jaeger401, line 16: I know it's not mentioned in the Eng. Portal doc, but do we discourage use of [`ready?`](https://graphql-ruby.org/mutations/mutation_authorization.html)? If not, should we accommodate that in here or in the resolve methods? We do have active usages of `ready?` in a few ez-rails mutations and I'd hate for someone to edit those and end up with confusing output from this cop.

lib/rubocop/cop/ezcater/graph_ql_type_file_order.rb
```diff
@@ -0,0 +1,112 @@
...
+            previous = MatchedNode.new(previous)
+            current = MatchedNode.new(current)
+
+            if previous.dsl_method? || current.dsl_method?
+              is_gql_type = true
+            end
+
+            next unless previous > current
+
+            msg = "#{previous.name} should be positioned after #{current.name}"
```
jaeger401, line 37: If I'm reading this right, it seems a little like a bubble sort. What happens if methods are misplaced beyond their nearest neighbor? e.g.:

```
argument
field
pundit_policy_class
pundit_role
description
```

Would a developer have to run several lint/fix cycles to get the code into the appropriate order? If so, perhaps it would be helpful to provide a reference to the Eng. Portal doc to help educate them?

lib/rubocop/cop/ezcater/graph_ql_type_file_order.rb
drapergeek: That's a great callout.

Given that we use Pundit I can't think of a reason to use `ready?` for any new code.

Given the current usage in ez-rails I think it's reasonable to add this to the ordering for now. If we as a team agree that `ready?` should not be used we can add another rule for that.

lib/rubocop/cop/ezcater/graph_ql_type_file_order.rb
drapergeek: > If I'm reading this right, it seems a little like a bubble sort. 

You are, it is. This became even more apparent when I started trying to get this to autocorrect.

> What happens if methods are misplaced beyond their nearest neighbor? 

They would have to run this multiple times. It's not ideal at all, hence the reason I was poking at the autocorrect.

I'm totally behind adding a link to the output. It would be great if all cops provided that kind of explanation!

PR #112, opened 2022-07-20
Add documentation for adding new cops and release existing cops

drapergeek:
## What did we change?
This enables a LOT of cops and cuts a new release with them enabled.

## Why are we doing this?

At some point we had a suggestion that new cops were to be added as `enabled: false` and then enabled on future major releases. This didn't work as planned and we had a lot of cops that were added and never enabled. To prevent this situation from happening again, I updated the documented guidance on new cops.

## How was it tested?
- [ ] Specs
- [ ] Locally


mikowitz (approved):
Makes sense to me! :shipit:

mikowitz:
Just make sure we bump the gem version and add a comment in the CHANGELOG!

PR #116, opened 2022-12-12
Update rubocop to 0.90.0

ransombriggs:
## What did we change?

I created a `2-x-stable` branch off of `2.4.0` tag to allow us to bump up ez-rails `rubocop` to get some fixes in for another gem that we would like to use to validate packwerk.   Basically `RuboCop::Cop::Base` was added in [0.8.7](https://github.com/rubyatscale/rubocop-packs/issues/47) but I had to bring it up to 0.9.0 to get a fix for [this issue](https://github.com/rubocop/rubocop/issues/8713).  Once I merge this back to `2-x-stable` I will cut a release from there.

## How was it tested?
- [ ] Specs
- [x] Locally


ransombriggs:
I manually created a version of this gem, installed locally and conservative updated rubocop and they did not require updates

PR #119, opened 2023-05-30
[DEM-266] Add `FeatureFlagValidName` cop

j-sm-n:
Jira ticket: [DEM-266](https://ezcater.atlassian.net/browse/DEM-266)

## What did we change?
Add `FeatureFlagValidName` cop to validate that feature flag string names match the same pattern constraints [that exist](https://github.com/ezcater/feature-flag-admin-rails/blob/main/app/models/concerns/naming_conventions.rb) in the Feature Flag Admin service.

## Why are we doing this?
[A similar cop](https://github.com/ezcater/ez-rails/blob/main/cops/rubocop/cops/feature_flags_flag_name.rb) exists locally in `ez-rails` already, but we wish to move this cop to the shared ezCater gem so that we can use Rubocop to check for invalid feature flag names before runtime instead of adding this check to the `stub_feature_flag` [test helper ](https://github.com/ezcater/ezcater_feature_flag-client-ruby/blob/b478aa2e51775fa3b3867495f7952fdf7a44e478/lib/ezcater_feature_flag/rspec.rb#L13)in the feature flag client gem.

See [DEVEX-2007](https://ezcater.atlassian.net/browse/DEVEX-2007) for more details.

## How was it tested?
- [x] Specs
- [ ] Locally

[DEVEX-2007]: https://ezcater.atlassian.net/browse/DEVEX-2007?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DEM-266]: https://ezcater.atlassian.net/browse/DEM-266?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ This gem is moving onto its own [Semantic Versioning](https://semver.org/) schem
 
 Prior to v1.0.0 this gem was versioned based on the `MAJOR`.`MINOR` version of RuboCop. The first release of the ezcater_rubocop gem was `v0.49.0`.
 
+## 5.3.0
+- Add `FeatureFlagNameValid` cop to validate correct feature flag name format, adopted from the cop in `ez-rails`: https://github.com/ezcater/ez-rails/blob/main/cops/rubocop/cops/feature_flags_flag_name.rb.
+
```
j-sm-n, line 11: I went with a minor version bump for now but I could also see an argument for a major version bump since this could require other squads to disable or actively add `TODO`'s through-out their codebases depending on how many feature flags they use.

spec/rubocop/cop/ezcater/feature_flag_name_valid_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+  describe "assignment to a constant" do
+    context "with a valid feature flag name" do
+      let(:line) { %[MY_FLAG = "Foo::Bar"] }
+
+      it "does not report an offense" do
+        expect(cop.offenses).to be_empty
+      end
+    end
+
+    context "with an invalid constant name" do
```
tristandunn, line 16: 🎱 To match the other contexts.

```suggestion
    context "with an invalid feature flag name" do
```

lib/rubocop/cop/ezcater/feature_flag_name_valid.rb
```diff
@@ -0,0 +1,91 @@
...
+              break
+            end
+          end
+
+          errors
+        end
+      end
+    end
+  end 
+end
```
tristandunn, line 91: Looks like the new file, this file and the test file, are missing new lines at the end. I thought there was a cop for this, but it may have been removed.

spec/rubocop/cop/ezcater/feature_flag_name_valid_spec.rb
j-sm-n: Thank you!

tristandunn (approved):

lib/rubocop/cop/ezcater/feature_flag_name_valid.rb
```diff
@@ -0,0 +1,91 @@
...
+        PATTERN
+
+        def_node_matcher :feature_flag_method_call, <<~PATTERN
+          (send
+            (_ _ {:EzFF :EzcaterFeatureFlag}) {:active? | :at_100? | :random_sample_active?} (str $_ ...))
+        PATTERN
+
+        def on_casgn(node)
+          feature_flag_constant_assignment(node) do |const_name, flag_name|
+            if const_name.end_with?("_FLAG")
```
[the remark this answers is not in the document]
j-sm-n: Thanks for calling that out--I think it makes sense to cover those cases as well. 👍🏼

lib/rubocop/cop/ezcater/feature_flag_name_valid.rb
```diff
@@ -0,0 +1,91 @@
...
+  module Cop
+    module Ezcater
+      # Match the naming validations in the feature flag service.
+      #
+      # @example
+      #
+      #   # good
+      #   EzFF.active?("Foo::Bar")
+      #   EzFF.active?("Foo::Bar::Baz")
+      #   MY_FLAG="Foo::Bar"; EzFF.active?(MY_FLAG)
```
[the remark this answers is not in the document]
j-sm-n: Ohh, good point. I'll add a test for that.

CHANGELOG.md
djcp: minor bumps are for new features so that tracks, and I believe most of our folks expect some rubocop slicing-and-dicing on upgrades.

spec/rubocop/cop/ezcater/feature_flag_name_valid_spec.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::Ezcater::FeatureFlagNameValid, :config do
+  subject(:cop) { described_class.new(config) }
+
+  before { inspect_source(line) }
+  describe "assignment to a constant" do
+    context "with a valid feature flag name" do
+      let(:line) { %[MY_FLAG = "Foo::Bar"] }
```
djcp, line 9: :8ball: It would be nice to see us testing a few more valid flag names here. Then our specs can serve as self documenting tests for what's a valid name.

spec/rubocop/cop/ezcater/feature_flag_name_valid_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+      it "does not report an offense" do
+        expect(cop.offenses).to be_empty
+      end
+    end
+
+    context "with an invalid feature flag name" do
+      let(:line) { %[MY_FLAG = "Foo:Bar"] }
```
djcp, line 17: :8ball: Same as above -it might be nice to test maybe a half dozen different "bad" flag names

djcp (approved):

CHANGELOG.md
j-sm-n: Ah I did just notice that it says this in the README:
```When adding a new cop, please enable the cop and release a new major version. This allows us to constantly roll out improvements without clients having their suite break unknowingly. 
```

So in the spirit of staying consistent, I'll change this to a major bump instead if that works with everyone.

PR #125, opened 2023-11-17
feat: Add Ezcater/GraphQL/NotAuthorizedScalarField cop, et al.

jmpage:
## What did we change?

This cop enforces the use of authorization for scalar GraphQL fields.

By default, it requires pundit and identifies all standard
graphql-ruby scalars as scalar values.

It can be configured to:
- Allow the "guard pattern" for authorization
- Identify additional types as scalars
- Ignore certain fields by name

### Development Changes

Adds a `new_cop` rake task which is helpful for generating new cops.

## Why are we doing this?

This cop can be helpful for finding GraphQL fields which lack adequate authorization. It only checks scalar fields, under the assumption that mutations and fields which reference object types should use more generalized authorization, which should be checked by future cops.

This cop is turned off by default so that application-owners can opt-in, minimizing disruption and providing an opportunity to identify and fix edge cases before making it mandatory.

This cop was the most complex one of the set to write as it needs to cover multiple mechanisms for handling authorization and multiple sets of syntax for defining resolvers. The other cops should be easier to write in the future and can likely follow this cop's example.

## How was it tested?
- [x] Specs
- [x] Locally

### DIFF
--- a/lib/rubocop/cop/ezcater/graphql/not_authorized_scalar_field.rb
+++ b/lib/rubocop/cop/ezcater/graphql/not_authorized_scalar_field.rb
@@ -1,0 +1,198 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Ezcater
+      module GraphQL
+        # Enforces authorization for scalar GraphQL fields. It
+        # identifies scalar fields by matching against the standard
+        # [graphql-ruby scalar types] by default.
+        #
+        # [graphql-ruby scalar types]: https://graphql-ruby.org/type_definitions/scalars.html
+        #
+        # @safety
+        #   This cop is unable to autocorrect violations as decision making is
+        #   required on the part of the developer.
+        #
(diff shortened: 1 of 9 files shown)


jeschneiderhan (approved):

lib/rubocop/cop/ezcater/graphql/not_authorized_scalar_field.rb
```diff
@@ -0,0 +1,198 @@
...
+        # @example AdditionalScalarTypes: ["UUID"]
+        #   # This example adds UUID to the types to observe.
+        #
+        #   # bad
+        #   field :secret_id, UUID
+        #
+        #   # good
+        #   field :secret_id, pundit_role: owner
+        #
+        # @example IgnoredFieldNames: [] (default)
```
drapergeek, line 79: What situation do you see this applying to?

drapergeek (approved):

PR #127, opened 2023-12-18
Add cop for time in expect

RitamBanerjee:
## What did we change?
We don't want people asserting on a call to `Time`
## Why are we doing this?
This leads to flaky time bomb tests
## How was it tested?
- [ ] Specs
- [ ] Locally


jmpage (commented):
Mind updating the CHANGELOG and version.rb as well as the README to note the existence of this cop?

maudineormsby:
Very nice!

jbeard6 (commented):
> I concur with Jen -- could you please update CHANGELOG.md, version.rb, and the README? Happy to approve once those are updated!

Ditto.

PR #128, opened 2024-04-02
CA-401 Lock rubocop_rspec below v2.28.x

genericsteele:
## What did we change?

This locks the rubocop_rspec gem below v2.28 to temporarily fix a regression caused by that issue.

## Why are we doing this?

[Slack](https://ezcater.slack.com/archives/CREPTH9LJ/p1712001323908889)

[rubocop_rspec v2.28.0](https://github.com/rubocop/rubocop-rspec/releases/tag/v2.28.0) was released with a regression that is causing our gems to throw "Ambiguous cop name `RSpec/Rails/HttpStatus`" errors: 

Screenshot 2024-04-02 at 12 44 30 PM

This temporary fix should unblock ezcater gem releases until a more permanent solution is available

## How was it tested?
- [ ] Specs
- [ ] Locally

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ This gem is moving onto its own [Semantic Versioning](https://semver.org/) schem
 
 Prior to v1.0.0 this gem was versioned based on the `MAJOR`.`MINOR` version of RuboCop. The first release of the ezcater_rubocop gem was `v0.49.0`.
 
+## 6.1.1
+
+- Lock `rubocop-rspec` below `v2.28.0` to avoid an upstream namespacing issue.
+
 ## 6.1.0
 
 - Add `Ezcater/GraphQL/NotAuthorizedScalarField` Cop which enforces the use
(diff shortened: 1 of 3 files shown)


jbeard6 (commented):
With the patched version [v2.28.2](https://github.com/rubocop/rubocop-rspec_rails/releases/tag/v2.28.2), I'm not sure this is necessary anymore. I haven't tested it, yet, but, if it works, would you have any objections to closing this in favor of the updated dependency?

genericsteele:
@jbeard6 The patched version is in a separate gem (rubocop-rspec_rails), so that would involve pulling in another dependency, and ensuring they are loaded correctly. I'd recommend rolling this change out first so we can unblock any gem releases, then following up with a more permanent fix.

mikowitz:
Do we have links to a couple of our workflows that show 
* the original error
* that this solution works

genericsteele:
Here's a link to a failing PR: https://github.com/ezcater/ezcater_accounting/actions/runs/8525581203/job/23354000587?pr=56

genericsteele:
@mikowitz Here's a passing fork of that failing PR with just the version of rubocop-rspec locked below v2.28: https://github.com/ezcater/ezcater_accounting/pull/57/checks

mikowitz (approved):
LGTM. Hopefully they'll fix this upstream and we can stop having to do this, but for now, let's get ourselves unblocked!

:shipit:

genericsteele:
@mikowitz I just bumped the version and updated CHANGELOG.md.

PR #129, opened 2024-09-20
fix: Disabling the cop that enforces variable name for Exceptions

carlosagp:
## What did we change?
Disabled the `Naming/RescuedExceptionsVariableName` cop

## Why are we doing this?
Single variable names were used heavily in the past due to screen constraints and memory sizes. Also because of math notations like `i, j, l` (remember C's `for (i = 1; i < 11; ++i)`? 
Ruby's greatest strength is readability, this change will allow devs to use a more significant name if desired and also keep `e` too for historical reasons, but mainly to not pigeon hole anyone to any specific variable name pattern.

## How was it tested?
- [ ] Specs
- [x] Locally

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ This gem is moving onto its own [Semantic Versioning](https://semver.org/) schem
 
 Prior to v1.0.0 this gem was versioned based on the `MAJOR`.`MINOR` version of RuboCop. The first release of the ezcater_rubocop gem was `v0.49.0`.
 
+## 6.1.2
+
+- Disable `Naming/RescuedExceptionsVariableName` so exception variable names are less restrictive.
+
+
 ## 6.1.1
 
 - Lock `rubocop-rspec` below `v2.28.0` to avoid an upstream namespacing issue.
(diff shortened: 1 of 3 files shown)


ransombriggs (approved):

drapergeek (approved):

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ This gem is moving onto its own [Semantic Versioning](https://semver.org/) schem
 
 Prior to v1.0.0 this gem was versioned based on the `MAJOR`.`MINOR` version of RuboCop. The first release of the ezcater_rubocop gem was `v0.49.0`.
 
+## 6.1.2
```
juliantrueflynn, line 9: I would think this is a minor version release and not a patch.

juliantrueflynn (approved):
Reposting what I mentioned in the [original Slack thread](https://ezcater.slack.com/archives/C94MF14JF/p1726849971982569):

> I personally prefer having `e`, and I don't see a reason it needs more specific naming, it's always going to be an exception being rescued and I would think the error class name defined gives insight on what kind of exception is rescued.
>
> Linting/formatting wise, I think it's a little too heavy handed, if it was removed I wouldn't care

ezveezus (approved):

PR #130, opened 2024-11-22
Drop support for ruby 2.6 & 3.0

ransombriggs:
## What did we change?

We tried to bump the ruby version in another gem and the `TargetRubyVersion` in conf/rubocop_gem.yml was blocking requiring a supported version of ruby.  Rather than override locally, it felt like a good time to drop support for unsupported versions of ruby.

## How was it tested?
- [ ] Specs
- [ ] Locally

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,4 +39,4 @@ workflows:
           matrix:
             parameters:
               # https://github.com/CircleCI-Public/cimg-ruby
-              ruby-version: ["2.7", "3.0", "3.1"]
+              ruby-version: ["3.1"]
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,4 +39,4 @@ workflows:
           matrix:
             parameters:
               # https://github.com/CircleCI-Public/cimg-ruby
-              ruby-version: ["3.1"]
+              ruby-version: ["3.1", "3.2", "3.3"]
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -39,4 +39,4 @@ workflows:
           matrix:
             parameters:
               # https://github.com/CircleCI-Public/cimg-ruby
-              ruby-version: ["2.7", "3.0", "3.1"]
+              ruby-version: ["3.1"]
```
gguerini, line 42: Should we also add `3.2` and `3.3` to the ruby matrix?

ransombriggs: Good idea, added in [2c8f5a3](https://github.com/ezcater/ezcater_rubocop/pull/130/commits/2c8f5a39bd7f725038f8fc7f6ecbec9a17587dd8)

gguerini (approved):

kevinmook (approved):

lib/ezcater_rubocop/version.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module EzcaterRubocop
-  VERSION = "6.1.2"
+  VERSION = "6.2.0"
```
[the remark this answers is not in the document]
ransombriggs: I changed to 7.0.0 in [8b91803](https://github.com/ezcater/ezcater_rubocop/pull/130/commits/8b918034e405c8eee9ce46572758896c09d98815), the only reason I picked minor bump is because that is what we did [last time](https://github.com/ezcater/ezcater_rubocop/pull/96/files), so I do not have strong opinions about this.

mikowitz (approved):
:shipit:

PR #133, opened 2024-12-12
Disable `BulkChangeTable` cop

tvec:
## What/Why did we change?

This config conflicts with [`strong_migrations`](https://github.com/ankane/strong_migrations) due to the usage of `change_table`. `strong_migrations` considers this to be unsafe. 

Rather than try to work around it, we are disabling this cop. More details [here]

[here]: https://github.com/ankane/strong_migrations/issues/204

## How was it tested?
- [ ] Specs
- [x] Locally

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ This gem is moving onto its own [Semantic Versioning](https://semver.org/) schem
 
 Prior to v1.0.0 this gem was versioned based on the `MAJOR`.`MINOR` version of RuboCop. The first release of the ezcater_rubocop gem was `v0.49.0`.
 
+## 7.1.1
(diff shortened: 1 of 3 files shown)


jmpage (approved):

tvec:
@jaeger401 good call! Just pushed up

phantasm66:
You have a whitespace that breaks our rubocop implementation: https://github.com/ezcater/ezcater_rubocop/blob/main/conf/rubocop_rails.yml#L27

tvec:
@phantasm66 Sorry about that! Let me fix that now

phantasm66:
> @phantasm66 Sorry about that! Let me fix that now

No worries!

PR #134, opened 2024-12-12
Remove stray space in BulkChangeTable  definition

tvec:
## What did we change?

Via #133, there was a stray whitespace in the config which caused:

```
Error: unrecognized cop or department Rails/ BulkChangeTable found in /usr/local/bundle/gems/ezcater_rubocop-7.1.1/conf/rubocop_rails.yml
Did you mean `Rails/BulkChangeTable`?
```

## How was it tested?
- [ ] Specs
- [x] Locally


phantasm66 (approved):
🎉

PR #135, opened 2024-12-16
Run simple check on Rubocop configs

tvec:
## What did we change?
- Add a simple script to validate configs within [`conf/`](https://github.com/ezcater/ezcater_rubocop/tree/main/conf)

## Why are we doing this?
- In #133, I had an invalid stray space but nothing told me about it within the PR and that lead to fun surprises

## Pics (Locally)
Screenshot 2024-12-16 at 11 09 40 AM

## Pics (Circle CI)
[Example Link
](https://app.circleci.com/pipelines/github/ezcater/ezcater_rubocop/193/workflows/782acca8-c1c3-4a50-bc3b-a3ed92a98807/jobs/697)
Screenshot 2024-12-16 at 11 23 54 AM

## How was it tested?
- [x] Specs
- [x] Locally

### DIFF
--- a/bin/check_configs.sh
+++ b/bin/check_configs.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+# Checks the validity of RuboCop config files in `conf/` directory (excluding the base config).
+
+CONFIGS=(
+  "conf/rubocop_gem.yml"
+  "conf/rubocop_rails.yml"
+)
+
+exit_code=0
+
+for config in "${CONFIGS[@]}"; do
+  if ! output=$(bundle exec rubocop --show-cops -c "$config" 2>&1); then
+    echo "❌ Error in $config"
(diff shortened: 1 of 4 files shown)


bin/check_configs.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+# Checks the validity of RuboCop config files in `conf/` directory (excluding the base config).
+
+CONFIGS=(
+  "conf/rubocop_gem.yml"
```
tvec, line 6: `rubocop.yml` was omitted from this list since it gets inherited by both these configs:

https://github.com/ezcater/ezcater_rubocop/blob/24b52dac3b57eef05ad76d4e42803273fa690829/conf/rubocop_gem.yml#L1-L2

PR #136, opened 2025-01-27
Add new cop to force bigint on foreign keys

bray:
## What did we change?

Add a new cop to force `bigint` on all (new) foreign keys. This includes safe auto-correction for some, but not all, cases. e.g. it will correct these:

- `t.integer :bar_id` -> `t.bigint :bar_id`
- `t.integer :bar_id, limit: <less than 8>` -> `t.bigint :bar_id`
- `add_reference :foos, :bars, type: :integer` -> `add_reference :foos, :bars`

but not these:

- `t.integer :bar_id, null: false, limit: <less than 8>`
- `add_reference :foos, :bars, type: :integer, index: { algorithm: :concurrently }`

There isn't an easy/good way to replace parts of a hash using the AST because of hanging commas and spaces. Even some internal methods weren't good.

But: this will still catch all these and similar cases, it just won't be able to auto-correct all of them.

## Why are we doing this?

This is a preventative measure in response to [this incident](https://ezcater.atlassian.net/wiki/spaces/BCIM/pages/4820762680/Incident-401+Critical+2024-12-12+-+Order+extras+primary+key+hit+maximum).

Requiring all new foreign keys to be `bigint` should help prevent FKs from running out of values before their referenced PKs.

[MX-895](https://ezcater.atlassian.net/browse/MX-895)

## How was it tested?
- [x] Specs
- [x] Locally - ran on a dummy migration in ez-rails

[MX-895]: https://ezcater.atlassian.net/browse/MX-895?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/rubocop/cop/ezcater/migration/bigint_foreign_key.rb
```diff
@@ -0,0 +1,264 @@
...
+        #     t.references :bar
+        #   end
+        #
+        #   # good
+        #   add_column :foos, :bar_id, :bigint
+        #
+        #   # good
+        #   add_reference :foos, :bar
+        #
+        class BigintForeignKey < Base # rubocop:disable Metrics/ClassLength
```
jbeard6, line 49: ⚠️ Warning: This is a driveby thought that may have already been addressed elsewhere.

This feels like a partial prevention as it only addresses the foreign key side of things. The primary issue is on the primary key side (e.g. `bars.id` rather than `bar_id`) and the foreign key is a side-effect of it. It does not matter if the FK is a `BIGINT` if the column that it references is an `INT` because the PK column will have exhausted its capacity long before the FK side.

We really need checks on both sides:
- That an ID column that will be populated by a sequence is a `BIGINT`, and
- That any FK columns that reference it are also a `BIGINT`.

Am I missing something? (It's a totally fair answer if there's a different unit of work for that.)

bray: For posterity: I just checked, and we upgraded ez-rails to Rails 5 in 2017. Of course, most of the core tables had already been created, and those are some of the most likely to run into this issue on the PK side. But PKs created in the last 7+ years are mostly safe.

It's the FKs that would continue to be an issue because we'll always be adding new FKs that reference big tables like orders.

lib/rubocop/cop/ezcater/migration/bigint_foreign_key.rb
```diff
@@ -0,0 +1,264 @@
...
+          extend AutoCorrector
+
+          MSG = <<~MSG.chomp
+            Use `bigint` instead of `integer` for foreign keys. This prevents them from potentially running out of int values before their referenced primary keys.
```
jmpage, line 53: Thoughts on this slight language change for clarity?:

```suggestion
            Use `bigint` instead of `integer` for foreign keys. This ensures that they are capable of storing values from referenced primary keys which are `bigint` or may eventually be migrated to `bigint`.
```

j-sm-n (approved):

PR #137, opened 2025-02-20
Pin maximum versions of `rubocop` and `rubocop-rails`

mikowitz:
## What did we change?

Pin maximum versions of `rubocop` and `rubocop-rails`

Also, include `Gemfile.lock` in version control. The recommendation to exclude this is outdated, and tools like Dependabot obviate the CI benefits of this approach that simultaneously make local development harder

## Why are we doing this?

Release `1.72` of `rubocop` begins using the `plugins` architecture, which is how `rubocop-rails` is included moving forward. This is not compatible with how we setup inheriting from this gem in our internal gems, and because this gem neither pins maximum versions, nor includes `Gemfile.lock` in version control, `bundle install` could install incompatible versions of those gems.

The goal here is to make version 8.1 of this gem compatible with the pre-`plugin` configuration for rubocop, and a subsequent major version will be compatible with the newer versions of the gems.

## How was it tested?
- [x] Specs
- [ ] Locally

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 /.bundle/
 /.yardoc
-/Gemfile.lock
 /_yardoc/
 /coverage/
 /doc/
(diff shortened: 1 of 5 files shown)


j-sm-n (approved):

j-sm-n:
I'm alllll in favor of checking in Gemfile.locks. Do we want to make this a recommendation for all of our gems moving forward? Or leave it up to squads? I'm cool with doing this for MX gems.

ransombriggs (approved):

PR #162, opened 2026-06-22
docs: deprecate ezcater_rubocop in favor of linter gem

j-sm-n:
## What did we change?
Added a deprecation / forwarding notice to the top of the README. `ezcater_rubocop` is superseded by the `linter` gem ([ezcater/linter-ruby](https://github.com/ezcater/linter-ruby)). The notice directs internal consumers to `linter` and tells external/public consumers to vendor their own RuboCop config (since `linter` is internal-only).

## Why are we doing this?
Part of retiring the public `ezcater_rubocop` repo now that internal consumers have migrated. This forwarding notice must land before the repo is archived (read-only).

Jira: https://ezcater.atlassian.net/browse/MX-1412

## How was it tested?
- [ ] Specs
- [x] Locally (README-only change; rendered notice reviewed)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,24 @@
+> [!IMPORTANT]
+> ## ⚠️ This gem is deprecated and this repository is archived
+>
+> `ezcater_rubocop` has been replaced by **`linter`** (repository:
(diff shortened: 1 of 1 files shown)


jbeard6 (approved):
🚁

